PR #1, opened 2023-03-27
Migrate Uno.Extensions.Template

dansiegel:

- n/a

## PR Type

- other

This migrates the Uno.Extensions.Template and the download combine logic for Uno.Templates from Uno.Studio


dansiegel:
We will make the updates to the template in a follow up PR to #2

PR #14, opened 2023-03-28
Mvvm or Mvux should only be added with Extensions Navigation

### LINKED ISSUE #12: When creating a new project using the vsix and selecting the blank template (without customization), the generated template includes the CommunityToolkit.Mvvm package installed. Is this by design?

dansiegel:

- closes #12

## PR Type

- Bugfix

CommunityToolkit or Reactive packages can be added to the project without a ViewModel/Model

## What is the new behavior?

These now require that Extensions Navigation be enabled as well.

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -758,12 +758,12 @@
     "useMvux": {
       "type": "computed",
       "datatype": "bool",
-      "value": "architectureEvaluator == 'mvux'"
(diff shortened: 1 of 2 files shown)


nickrandolph:
I think this should be changed to set architecture to mvvm when blank is selected
https://github.com/unoplatform/uno.templates/blob/0216cb0dec67e59ea753ef603a1dbee89b372d9d/src/Uno.Templates/content/unoapp/.template.config/template.json#L745

nickrandolph (approved):

PR #21, opened 2023-03-29
Dependencies

### LINKED ISSUE #20: Lang Version should not be included when not using net6.0
## Current behavior

The LangVersion is always specified and is calculated based on the TFM. Currently does not work with net8

## Expected behavior

We only need to specify the LangVersion with net6.0 where we have the need to use netstandard for the Skia / Wasm targets. With net7+ we do not need to specify the language version as the default is correct for these.

dansiegel:

- closes #20 

## PR Type

- Bugfix
- Dependencies

References old dependencies.
Specifies LangVersion all of the time

## What is the new behavior?

LangVersion only included in the Build Props for net6.0
Windows Sdk BuildTools updated
net8-pre1 packages updated to pre2
Uno Bootstrap updated to target v8 by default when generating a net8.0 project.


src/Uno.Templates/content/unoapp/.template.config/template.json
```diff
@@ -498,8 +498,34 @@
...
+          }
+        ]
+      }
+    },
+    "unoWasmBootstrapVersionEvaluator": {
+      "type": "generated",
+      "generator": "coalesce",
+      "parameters": {
+        "sourceVariableName": "unoWasmBootstrapVersion",
+        "fallbackVariableName": "unoWasmBootstrapVersionDefault"
```
nickrandolph, line 527: This still doesn't protect against invalid bootstrap versions eg specifying stable 7.x bootstrapper when tfm set to net8

nickrandolph (approved):

PR #66, opened 2023-04-12
fix: Reordering properties, methods etc

nickrandolph:
## PR Type

- Bugfix

Properties/Fields/Methods in wrong order

## What is the new behavior?

Correct order

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginViewModel.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginViewModel.cs
@@ -1,3 +1,4 @@
+//-:cnd:noEmit
 using System.Windows.Input;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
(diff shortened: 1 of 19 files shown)


src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginViewModel.cs
```diff
@@ -6,18 +7,16 @@ namespace MyExtensionsApp._1.Presentation;
...
 	[ObservableProperty]
 	private string? _username;
 
 	[ObservableProperty]
 	private string? _password;
 
-	public ICommand Login { get; }
-
-//+:cnd:noEmit
-#if useLocalization
```
jeromelaban, line 20: Not needed anymore?

jeromelaban (approved):

src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginViewModel.cs
nickrandolph: yep not needed, was accidentally left in

PR #74, opened 2023-04-13
feat: Include additional debug target

jeromelaban:


spouliot (approved):

src/Uno.Templates/content/unoapp/.vscode/launch.json
```diff
@@ -41,6 +41,13 @@
       // For more information about the 'console' field, see https://aka.ms/VSCode-CS-LaunchJson-Console
       "console": "internalConsole",
       "stopAtEntry": false
-    }
+    },
+    {
+      "name": "Uno Platform Mobile",
```
dansiegel, line 46: Should this say Mobile when it appears to be for Android?

PR #93, opened 2023-05-01
fix: Correct Background of MainPage for Blank preset with Material

nickrandolph:


src/Uno.Templates/content/unoapp/MyExtensionsApp._1/MainPage.xaml.cs
```diff
@@ -7,7 +7,8 @@ public MainPage()
 	{
 //+:cnd:noEmit
 #if useCsharpMarkup
-		this.Content(new StackPanel()
+		this.Background(Theme.Brushes.Background.Default)
```
dansiegel, line 10: We need to update where we have `Theme.Brushes`... as it only works for Material

nickrandolph: Updated

dansiegel (approved):

PR #118, opened 2023-05-12
Rename Window as public property

### LINKED ISSUE #120: App should contain a MainWindow property rather than a private _window field
See: https://github.com/unoplatform/uno/blob/57248d0a6712a8f3796ad680ad4fa9f9f23b87f4/doc/articles/features/windows-storage-pickers.md#L54-L60

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

dansiegel:

- closes #120

## PR Type

- Refactoring 

The visibility of the `_window` property is inconsistent depending on whether you are creating a blank or recommended app.

## What is the new behavior?

The visibility is normalized as public with a private setter and the property has been renamed to be consistent with coding standard for a public property.

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
@@ -3,7 +3,7 @@ namespace MyExtensionsApp._1;
 
 public class App : Application
 {
-	public static Window? MainWindow { get; private set; }
+	protected Window? MainWindow { get; private set; }
 
 	protected override void OnLaunched(LaunchActivatedEventArgs args)
 	{
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
@@ -3,25 +3,25 @@ namespace MyExtensionsApp._1;
 
 public class App : Application
 {
-	public static Window? AppWindow { get; private set; }
+	public static Window? MainWindow { get; private set; }
 
 	protected override void OnLaunched(LaunchActivatedEventArgs args)
 	{
 #if NET6_0_OR_GREATER && WINDOWS && !HAS_UNO
-		AppWindow = new Window();
+		MainWindow = new Window();
 #else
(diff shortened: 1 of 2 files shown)


src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
```diff
@@ -3,25 +3,25 @@ namespace MyExtensionsApp._1;
 
 public class App : Application
 {
-	public static Window? _window;
+	public static Window? AppWindow { get; private set; }
```
nickrandolph, line 6: Why is this public?
Why is this a static?

nickrandolph:
My preference would be to use MainWindow instead of AppWindow (see this PR: https://github.com/unoplatform/uno.templates/pull/108)
- this is a property within the App class, so "App" prefix is redundant
- Main prefix pays tribute to the fact that it is the first/primary/main window of the app - even if you then create subsequent windows

MartinZikmund:
I think the `AppWindow` naming could be also confused with https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.appwindow?view=windows-app-sdk-1.2

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
MartinZikmund: I think it is because making it `public static` makes it a bit easier to access - because `Application.Current` is just a plain `Microsoft.UI.Xaml.Application` so you would have to do `((App)Application.Current).MainWindow`, this way you can do just `App.MainWindow`.

nickrandolph (approved):

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.blank.cs
```diff
@@ -3,25 +3,25 @@ namespace MyExtensionsApp._1;
 
 public class App : Application
 {
-	public static Window? _window;
+	protected Window? MainWindow { get; private set; }
```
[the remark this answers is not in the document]
dansiegel: Even better expose a service that is registered in DI that you could make use of

PR #122, opened 2023-05-12
Remove nowarn for xml comments

### LINKED ISSUE #38: CS1591 should not be disabled
https://github.com/unoplatform/uno.templates/blob/9e09f5861948060dfbd9c11c5f24e7fc6ef115e2/src/Uno.Templates/content/unoapp/MyExtensionsApp.1.Server/MyExtensionsApp.1.Server.csproj#L6

https://github.com/unoplatform/uno.templates/blob/9e09f5861948060dfbd9c11c5f24e7fc6ef115e2/src/Uno.Templates/content/unoapp/MyExtensionsApp.1.DataContracts/MyExtensionsApp.1.DataContracts.csproj#L6

We should do one of the following:

- Not setting `GenerateDocumentationFile` to true, and remove the NoWarn of CS1591
- Keeping `GenerateDocumentationFile` as true, *address* the warnings, and remove the NoWarn of CS1591

Compiler warnings shouldn't be suppressed in templates unless absolutely necessary.

It's best if we can get rid of **all** the NoWarn, but for now at least let's remove CS1591.

dansiegel:

- closes #38 

## PR Type

- Refactoring

We disable CS1591 in the DataContracts and Server so that we do not annoy developers who are not adding XML docs to everything while enabling full swagger docs from the generated template

## What is the new behavior?

Making the minimal API extensions internal to avoid issues with requiring XML docs, and adding a comment to the Serialization Context in the DataContracts. The NoWarn has been removed from both projects.

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp._1.DataContracts/MyExtensionsApp._1.DataContracts.csproj
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp._1.DataContracts/MyExtensionsApp._1.DataContracts.csproj
@@ -2,8 +2,6 @@
 	<PropertyGroup>
 		<TargetFramework>$libraryBaseTargetFramework$</TargetFramework>
 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
-		<!-- Suppress Compiler warnings for elements without XML Docs -->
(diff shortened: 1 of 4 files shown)


Youssef1313 (approved):
LGTM!

PR #129, opened 2023-05-17
feat: Adding default logging

nickrandolph:


nickrandolph:
@dansiegel I think somewhere along the line we've lost uno logging output. We should test this to make sure that using the different log levels we see uno output

nickrandolph:
This PR is ok to proceed. There's an issue in Logging for wasm/skia but not related to the template (see https://github.com/unoplatform/uno.extensions/pull/1477)

PR #141, opened 2023-05-29
feat: Adding script files for launching

nickrandolph:
## PR Type

What kind of change does this PR introduce?
- Feature

## What is the current behavior?

No way to restrict which tfms are loaded and built when opening solution

## What is the new behavior?

Solution filters for Mobile, Wasm, Gtk and Windows
OverrideTargetFrameworks can be specified as environmental (or msbuild) value to limit which targets are built for the class library
MobileOverrideTargetFrameworks is required to limit the tfms for the mobile project
debug-xxxx.bat script files are available for setting environmental variables, clearing obj/bin folders, doing a restore and launching the appropriate solution filter.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [ ] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #142, opened 2023-05-29
Enable single target builds

dansiegel:

- Replaces #141

## PR Type

- Feature

The projects core "shared" project and the mobile project will multi-target as necessary.

## What is the new behavior?

Developers may now optionally uncomment an import line in the Directory.Build.props and rename the sample file to single target. This is an opt in feature. We additionally have added the official VisualStudio .gitignore with the addition of a line to ignore the solution-config.props so that if a developer chooses to use this they will not accidently check the file into source control.

It will be up to the developer to create any solution filters that they may want. The linked docs explain this process in more detail.

### DIFF
--- a/src/Uno.Templates/Uno.Templates.csproj
+++ b/src/Uno.Templates/Uno.Templates.csproj
@@ -47,6 +47,15 @@
 
 		<RemoveDir Directories="$(TemplateContentFolder)" />
 
+		<!-- Ensure we keep the .gitignore up to date with the latest changes to the official VisualStudio.gitignore -->
+		<DownloadFile
+			DestinationFolder="$(TemplateContentFolder)/unoapp" 
+			DestinationFileName=".gitignore" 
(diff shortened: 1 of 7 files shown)


src/Uno.Templates/Uno.Templates.csproj
```diff
@@ -47,6 +47,15 @@
 
 		<RemoveDir Directories="$(TemplateContentFolder)" />
 
+		<!-- Ensure we keep the .gitignore up to date with the latest changes to the official VisualStudio.gitignore -->
+		<DownloadFile
```
nickrandolph, line 51: Can we move this to a different PR  - this is something that can go forward regardless

dansiegel: now that we're moving this to ready for review this should be fine to leave here

nickrandolph (approved):

PR #217, opened 2023-08-11
Normalizing whitespace & indents

dansiegel:

- n/a

## PR Type

- Refactoring

Whitespace uses mix of spaces and tabs. The default version of windows 10 in the overrides does not match the version of Windows 10 that we are targeting

## What is the new behavior?

Whitespaces & indents have been normalized, and the overrides file targets the correct version of Windows 10.


nickrandolph:
@mergify backport legacy/4x

PR #262, opened 2023-09-01
feat: version bump and enabling maui validation

nickrandolph:
- Increments Extensions version
- Adds back the .NET MAUI Validation steps

### DIFF
--- a/build/templates/package-validation.yml
+++ b/build/templates/package-validation.yml
@@ -7,15 +7,22 @@ steps:
 - checkout: self
   clean: 'true'
 
-- ${{ if contains(parameters.arguments, 'net8.0') }}:
-  - template: dotnet-install-windows.yml
-    parameters:
-      DotNetVersion: 8.0.100-preview.4.23260.5
-      UnoCheck_Version: 1.13.0-dev.7
-      UnoCheck_Manifest: https://raw.githubusercontent.com/unoplatform/uno.check/d26f9772669da46e1fb0db6ad205ff30842ab5ef/manifests/uno.ui-preview-major.manifest.json
+- powershell: |
+    $templateArgs = '${{ parameters.arguments }}'
+    Write-Host "TemplateArgs = '$templateArgs'"
+    $dotnetVersion = '7.0.306'
 
-- ${{ if not(contains(parameters.arguments, 'net8.0')) }}:
-  - template: dotnet-install-windows.yml
+    if ($templateArgs.Contains('net8')) {
+      $dotnetVersion = '8.0.100-preview.7.23376.3'
+    }
+
+    Write-Host "DotNetVersion = $dotnetVersion"
+    Write-Output "##vso[task.setvariable variable=DotNetVersion]$dotnetVersion"
+  displayName: Evaluate Template Args
+
+- template: dotnet-install-windows.yml
+  parameters:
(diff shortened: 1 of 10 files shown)


nickrandolph:
@mergify backport legacy/4x release/stable/4.10

build/templates/package-validation.yml
```diff
@@ -47,7 +48,7 @@ steps:
       $ErrorActionPreference = 'Stop'
 
       # Create app with defaults
-      dotnet new unoapp -o UnoApp1 ${{ parameters.arguments }}
+      dotnet new unoapp -skip -o UnoApp1 ${{ parameters.arguments }}
```
Youssef1313, line 51: What does the `-skip` flag do?

src/Uno.Templates/content/unoapp/Directory.Build.props
```diff
@@ -62,6 +68,7 @@
 				<SupportedOSPlatformVersion>10.0.18362.0</SupportedOSPlatformVersion>
 				<TargetPlatformMinVersion>10.0.18362.0</TargetPlatformMinVersion>
 				<RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
+				<EnableCoreMrtTooling Condition=" '$(BuildingInsideVisualStudio)' != 'true' ">false</EnableCoreMrtTooling>
```
Youssef1313, line 71: This is to allow `dotnet build` to work for WinUI from command-line?

Would like @jeromelaban to confirm if there are any concerns with this.

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/MyExtensionsApp._1.csproj
```diff
@@ -107,6 +107,10 @@
 		<!--#if (useMauiPackageReference)-->
 		<PackageReference Include="Microsoft.Maui.Controls" />
 		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" />
+		<!--#else -->
+		<!--#if (mauiEmbedding)-->
+		<PackageReference Include="Microsoft.Maui.Graphics" VersionOverride="$(MauiVersion)" />
```
Youssef1313, line 112: Shouldn't we have a single version of Maui in Directory.Packages.props, and not need to have any kind of version overrides?

Youssef1313 (approved):

build/templates/package-validation.yml
nickrandolph: Avoids doing a restore as part of creating template - this makes the whole process quicker since the dotnet build does a restore anyhow

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/MyExtensionsApp._1.csproj
nickrandolph: If only - this is a workaround because Maui uses sdk/workload version for various libraries

src/Uno.Templates/content/unoapp/Directory.Build.props
nickrandolph: Correct, that's what this is for. Otherwise have to use msbuild

PR #285, opened 2023-09-14
fix: removed toolkit xlmns

pictos:
Just a quick fix to remove the xmlns for CommunityToolkit.Maui from our template


nickrandolph (approved):

nickrandolph:
@mergify backport legacy/4x release/stable/4.10

PR #336, opened 2023-10-10
feat: Add flag to enable analyzers

### LINKED ISSUE #243: Enable more analyzers and make sure they are run on templates
We probably should be adding template tests that also run with analyzers enabled so we catch all those issues. Could you take a look?

_Originally posted by @jeromelaban in https://github.com/unoplatform/uno/issues/11291#issuecomment-1428265537_

Youssef1313:
GitHub Issue (If applicable): closes #243

## PR Type

- Feature

Analyzers are not enabled.

## What is the new behavior?

Analyzers can be enabled via an option that defaults to true.

## PR Checklist

- [ ] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [ ] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)

## Other information

### DIFF
--- a/src/Uno.Templates/content/unoapp/.globalconfig
+++ b/src/Uno.Templates/content/unoapp/.globalconfig
@@ -1,0 +1,4 @@
+is_global = true
+
+# IDE0055: Fix formatting
+dotnet_diagnostic.IDE0055.severity = warning
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -406,6 +406,13 @@
       "type": "parameter",
       "datatype": "bool"
     },
+    "analyzers": {
+      "displayName": "Analyzers",
+      "description": "Enable C# analyzers",
+      "type": "parameter",
+      "datatype": "bool",
+      "defaultValue": "true"
+    },
     "configuration": {
       "displayName": "Configuration",
       "description": "Load configuration information from appsettings.json",
--- a/src/Uno.Templates/content/unoapp/Directory.Build.props
+++ b/src/Uno.Templates/content/unoapp/Directory.Build.props
@@ -52,6 +52,13 @@
     <AndroidMaterialVersion  Condition=" '$(AndroidMaterialVersion)' == '' ">1.9.0.2</AndroidMaterialVersion>
     <AndroidXNavigationVersion  Condition=" '$(AndroidXNavigationVersion)' == '' ">2.6.0.1</AndroidXNavigationVersion>
     <!--#endif-->
+    <!--#if (analyzers)-->
+
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>Recommended</AnalysisMode>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <!--#endif-->
   </PropertyGroup>
 
   <Choose>
--- a/src/Uno.Templates/content/unoapp/Directory.Build.props
+++ b/src/Uno.Templates/content/unoapp/Directory.Build.props
@@ -58,6 +58,10 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <AnalysisMode>Recommended</AnalysisMode>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+
+    <!-- CA1010: Type 'MainPage' directly or indirectly inherits 'IEnumerable' without implementing 'IEnumerable<T>' -->
+    <!-- CA1848: For improved performance, use the LoggerMessage delegates -->
+    <NoWarn>$(NoWarn);CA1010;CA1848</NoWarn>
     <!--#endif-->
   </PropertyGroup>
 
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.recommended.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.recommended.cs
@@ -1,4 +1,6 @@
 //-:cnd:noEmit
+using System.Globalization;
+
 namespace MyExtensionsApp._1;
 
 //+:cnd:noEmit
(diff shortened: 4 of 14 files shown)


src/Uno.Templates/content/unoapp/Directory.Build.props
```diff
@@ -52,6 +52,17 @@
...
     <!--#endif-->
+    <!--#if (analyzers)-->
+
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>Recommended</AnalysisMode>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+
+    <!-- CA1010: Type 'MainPage' directly or indirectly inherits 'IEnumerable' without implementing 'IEnumerable<T>' -->
+    <!-- CA1848: For improved performance, use the LoggerMessage delegates -->
```
Youssef1313, line 63: We will need to look into this.

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/App.recommended.cs
```diff
@@ -123,7 +125,7 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
                         credentials ??= new Dictionary<string, string>();
                         credentials[TokenCacheExtensions.AccessTokenKey] = "SampleToken";
                         credentials[TokenCacheExtensions.RefreshTokenKey] = "RefreshToken";
-                        credentials["Expiry"] = DateTime.Now.AddMinutes(5).ToString("g");
+                        credentials["Expiry"] = DateTime.Now.AddMinutes(5).ToString("g", DateTimeFormatInfo.InvariantInfo);
```
Youssef1313, line 128: NOTE: This is a behavioral change. When using a null provider, the used default will be CurrentInfo which doesn't seem suitable here.

https://github.com/dotnet/runtime/blob/4001e074ee648506d64277729a7cce089a4996e2/src/libraries/System.Private.CoreLib/src/System/Globalization/DateTimeFormatInfo.cs#L314

src/Uno.Templates/content/unoapp/Directory.Build.props
```diff
@@ -52,6 +52,17 @@
     <AndroidMaterialVersion  Condition=" '$(AndroidMaterialVersion)' == '' ">1.9.0.2</AndroidMaterialVersion>
     <AndroidXNavigationVersion  Condition=" '$(AndroidXNavigationVersion)' == '' ">2.6.0.1</AndroidXNavigationVersion>
     <!--#endif-->
+    <!--#if (analyzers)-->
+
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>Recommended</AnalysisMode>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Youssef1313, line 60: We probably should remove this, and instead pass it here:

https://github.com/unoplatform/uno.templates/blob/c098cae34ce35b9025f7762eaa81691b5f291f93/build/templates/package-validation.yml#L79

nickrandolph:
@Youssef1313 why are we making this an option in the template/wizard? Should we just enable this by default

Also, can you fill in the PR template

Youssef1313:
@nickrandolph I'm okay making this not optional. But I thought it would be good to give the user a way to opt-out from including the analyzers.

jeromelaban:
Note that analyzers are generally a pain to use for developers during debug session, particularly when starting to use a framework and/or language. Let's not make the developer's experience difficult.

Youssef1313:
@jeromelaban So is the suggestion to keep the option off by default?

nickrandolph:
The fall back would be to leave this code there but commented out with a note explaining why it's good practice to enable them

Youssef1313:
> The fall back would be to leave this code there but commented out with a note explaining why it's good practice to enable them

That makes it harder to validate in CI

src/Uno.Templates/content/unoapp/.globalconfig
```diff
@@ -0,0 +1,4 @@
+is_global = true
+
+# IDE0055: Fix formatting
+dotnet_diagnostic.IDE0055.severity = warning
```
dansiegel, line 4: I'd rather not add this file. You should be able to set the diagnostic in the .editorconfig

src/Uno.Templates/content/unoapp/MyExtensionsApp._1/Services/Caching/WeatherCache.cs
```diff
@@ -38,7 +38,7 @@ public async ValueTask<IImmutableList<WeatherForecast>> GetForecast(Cancellation
 #if (useLogging)
             _logger.LogWarning("App is offline and cannot connect to the API.");
 #endif
-            throw new Exception("No internet connection");
+            throw new InvalidOperationException("No internet connection");
```
dansiegel, line 41: Thinking maybe change this to:
```suggestion
            throw new NetworkException(NetworkExceptionCategory.Offline, 0, "No internet connection");
```

src/Uno.Templates/content/unoapp/.template.config/template.json
```diff
@@ -406,6 +406,13 @@
       "type": "parameter",
       "datatype": "bool"
     },
+    "analyzers": {
+      "displayName": "Analyzers",
+      "description": "Enable C# analyzers",
```
dansiegel, line 411: Let's give this a better description... what do I get with these analyzers? Technically even without this Analyzers are enabled in the project... it's a specific subset that we're enabling here...

src/Uno.Templates/content/unoapp/.globalconfig
Youssef1313: .globalconfig is now considered the "modern"/"encouraged" alternative of .editorconfig. I'd definitely use .globalconfigs for setting severities instead of .editorconfigs. It also has a "slight" performance improvement on Roslyn side since globalconfigs are evaluated per project rather than per file.

dansiegel: Unless we have something compelling from Microsoft for why we should move away from .editorconfig I'm hesitant to allow yet another file in. Not to mention what you are doing here is already supported in .editorconfig, and we already configure diagnostics in the .editorconfig. We don't need to make it harder on developers by giving them 2 files to maintain.

dansiegel:
@Youssef1313 closing this as stale. Feel free to update this and reopen.

PR #427, opened 2023-11-10
Control .NET version globally from Directory.Build.props

dansiegel:

- n/a

## PR Type

- Feature

If you create the app with net7 and want to experiment with net8 you need to find/replace net7 with net8 in all of the projects.

## What is the new behavior?

You can now simply change the `DotNetVersion` in the Directory.Build.props. This will globally update the entire solution to use the version that you want. 

**NOTE:** This may result in undesired behavior such as with MAUI Embedding where get the requirement for the additional NuGet Packages, and you may have API breaks in libraries you are referencing when changing from net7 to net8 to net9.

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -1455,60 +1455,32 @@
         "datatype": "string",
         "cases": [
           {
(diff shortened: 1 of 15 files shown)


nickrandolph (approved):

nickrandolph:
@mergifyio backport release/stable/5.0

PR #446, opened 2023-11-16
fix: Renaming login and fixing url

nickrandolph:
Minor fixes for issues encountered during .NET Conf session
- Loginpage.xaml.cs doesn't get renamed to LoginPage.cs when using c# markup
- url in appsettings has wrong port


nickrandolph:
@mergifyio backport release/stable/5.0

dansiegel (approved):

PR #510, opened 2024-01-16
Use Uno.Sdk

### LINKED ISSUE #365: Why does Mobile project need to be reloaded on first solution open?
## Current behavior

We now get this everytime a Uno app solution is opened for the first time/created. Can we prevent it? It does not happen with MAUI, so it must be Uno specific

image

## Expected behavior

No error

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

dansiegel:

- closes #365

## PR Type

- Bugfix
- Feature

Projects use the Microsoft.NET.Sdk. The Mobile head lacks the necessary project capabilities which causes Visual Studio on Windows to want to reload the project.

## What is the new behavior?

Projects now use the Uno.Sdk. NuGet packages that ship as part of the Uno repo are now versioned with the $(UnoVersion) from the Uno.Sdk. A global.json has been added to provide an easier way to manage and the version of Uno globally across the solution.

## PR Checklist

### DIFF
--- a/src/Uno.Templates/content/unoapp/global.json
+++ b/src/Uno.Templates/content/unoapp/global.json
@@ -1,0 +1,6 @@
+{
+  "msbuild-sdks": {
+    "Uno.Sdk": "$UnoWinUIVersion$",
+    "Microsoft.Build.NoTargets": "3.7.56"
+  }
+}
--- a/src/Uno.Templates/Uno.Templates.csproj
+++ b/src/Uno.Templates/Uno.Templates.csproj
@@ -19,7 +19,7 @@
 	<PropertyGroup>
 		<!-- If you update Extensions or Uno version, make sure you update the versions in the reinstall.ps1 file too !-->
 		<UnoExtensionsVersion Condition="'$(UnoExtensionsVersion)' == ''">4.0.0-dev.212</UnoExtensionsVersion>
-		<UnoVersion Condition="'$(UnoVersion)' == ''">5.1.0-dev.1459</UnoVersion>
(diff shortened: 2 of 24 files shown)


src/Uno.Templates/content/unoapp/global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "msbuild-sdks": {
+    "Uno.Sdk": "$UnoWinUIVersion$",
```
nickrandolph, line 3:
```suggestion
    "Uno.Sdk": "$UnoVersion$",
```
Should this be `$UnoVersion$` ?

dansiegel: No that isn't what the Template uses to update the version

nickrandolph:
I definitely think that we need comments/readme that indicates where $UnoVerison comes from
I'd also consider having variables defined for Extensions, Toolkit, Theme, Resizetizer packages. I would make sense for the sdk to be moved into this repo perhaps. Versioning could still align with Uno but has the advantage of making it easier for managing versioning of other pieces of uno platform. It would also mostly negate the need for creating meta-package

dansiegel:
> I definitely think that we need comments/readme that indicates where $UnoVerison comes from
Right... but where do you feel is the right place for this? Should we add a ReadMe to the Main Project?

jeromelaban (approved):

PR #513, opened 2024-01-24
fix: adding intentional break for invalid project name

### LINKED ISSUE #474: Character restrictions are not enforced via the commandline
## Current behavior

When creating a project from the templates in Visual Studio, there is a check for which characters are acceptable as part of the project (or solution) name

There's no equivalent test for the project name when creating a solution from the command line.
Is this an oversight? 

This means it's possible to create apps that won't build, and there's no obvious indication of what's wrong from the error messages that are produced.
I understand that limiting the characters accepted avoids the need to deal with safe project names and escape characters, but I wanted to draw attention to the fact that these custom rules aren't being applied consistently when projects are being created outside of VS. 

## Expected behavior

Characters that are valid as part of a project name can be used, regardless of whether the wizard is used or the templates are used from the command line.

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

The equals sign is technically valid as part of a project name but currently breaks on .net 8 due to https://github.com/dotnet/roslyn/issues/71022

dansiegel:

- closes #474

## PR Type

From the CLI you are able to create a project with a space in the Project Name

## What is the new behavior?

If you create a project with a space in the project name a target will be added to the Directory.Build.targets. This will create an error to let you know you've created a project with an invalid name and that you should recreate the project without the use of spaces in the name.

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -83,6 +83,15 @@
       },
       "replaces": "$sanitizedProjectName$"
     },
+    "failBuildOnInvalidProjectName": {
+      "type": "generated",
+      "generator": "regexMatch",
+      "datatype": "bool",
+      "parameters": {
+        "source": "name",
+        "pattern": "\\s"
+      }
+    },
     "appId": {
       "displayName": "Application ID",
       "type": "parameter",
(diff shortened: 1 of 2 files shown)


jeromelaban (approved):

jeromelaban:
@mergifyio backport release/stable/5.1

PR #516, opened 2024-01-24
Remove unused Logging Condition

### LINKED ISSUE #382: Logging is not enabled in WinAppSDK head
`NETFX_CORE` is not defined in WinAppSDK, causing the logging to be effectively disabled:

https://github.com/unoplatform/uno.templates/blob/a4ffbe5cdac9e2ee23933611489b82f228ec2000/src/Uno.Templates/content/unoapp/MyExtensionsApp._1.Base/AppHead.xaml.cs#L65

dansiegel:

- closes #382

## PR Type

- Bugfix

The template includes an unused condition for NETFX_CORE which came from the legacy template but is not defined. 

## What is the new behavior?

As WinAppSdk already uses the Console Logger there is no reason to switch this to the Debug logger by default. As a result we are simply removing the `NETFX_CORE` condition with the Debug Logger


jeromelaban (approved):

jeromelaban:
@mergifyio backport release/stable/5.1

PR #590, opened 2024-03-20
fix: fix building single project from linux

spouliot:

## PR Type

- Bugfix

When building under Linux `dotnet restore` fails because `ios` and `maccatalyst` workloads are **not** available on that platform.

## What is the new behavior?

`netX.Y-ios` and `netX.y-maccatalyst` are removed from the `<TargetFrameworks>` on Linux solving the `dotnet restore` issue.

## PR Checklist

## Other information

https://github.com/dotnet/runtime/issues/85505

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MyExtensionsApp.1.csproj
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MyExtensionsApp.1.csproj
@@ -1,11 +1,14 @@
 <Project Sdk="Uno.Sdk">
   <PropertyGroup>
+    <TargetFrameworks>$unoTargetFrameworks$;$(DotNetVersion)</TargetFrameworks>
     <!--#if (useWinAppSdk) -->
-    <TargetFrameworks>$unoTargetFrameworks$;$(DotNetVersion)</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) or '$(EnableWindowsTargeting)' == 'true'">$(TargetFrameworks);$(DotNetVersion)-windows10.0.19041</TargetFrameworks>
-    <!--#else -->
-    <TargetFrameworks>$unoTargetFrameworks$;$(DotNetVersion)</TargetFrameworks>
     <!--#endif -->
+    <!--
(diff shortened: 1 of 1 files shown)


src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MyExtensionsApp.1.csproj
```diff
@@ -1,11 +1,14 @@
...
+    <!--#if (useWinAppSdk) -->
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) or '$(EnableWindowsTargeting)' == 'true'">$(TargetFrameworks);$(DotNetVersion)-windows10.0.19041</TargetFrameworks>
-    <!--#else -->
-    <TargetFrameworks>$unoTargetFrameworks$;$(DotNetVersion)</TargetFrameworks>
     <!--#endif -->
+    <!--
+      Under Linux we can't restore ios or maccatalyst workloads are not available on linux https://github.com/dotnet/runtime/issues/85505
+      dotnet restore would fail if they are not removed from the list
+    -->
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('linux')) == 'true'">$(TargetFrameworks.Replace("$(DotNetVersion)-ios","").Replace("$(DotNetVersion)-maccatalyst",""))</TargetFrameworks>
```
nickrandolph, line 11: Since the template generates the list of TFS (see $unoTargetFrameworks$) rather than doing a replace here we should just list only the TFS supported by linux.

dansiegel: I'd really rather avoid putting this in the templates altogether

jeromelaban: After a discussion with @dansiegel, we're going to move this particular behavior to the SDK, and remove the non-supported frameworks in a target. That will simplify the template significantly.

dansiegel:
> we're going to move this particular behavior to the SDK, and remove the non-supported frameworks in a target

Closing as this behavior is going to be taken care of in the Sdk.

spouliot:
definitively nicer!

PR #591, opened 2024-03-22
Misc updates

jeromelaban:


src/Uno.Templates/content/unoapp/.vscode/launch.json
```diff
@@ -21,15 +21,15 @@
       "name": "Debug (Chrome, WebAssembly)",
       "type": "chrome",
       "request": "launch",
-      "url": "http://localhost:5001",
+      "url": "http://localhost:5000",
```
dansiegel, line 24: What about the launchsettings

PR #597, opened 2024-03-27
refactor: Restore original targetframework

jeromelaban:
This change restores the original netX.0 target framework in csproj, and removs the use of `$(DotnetVersion)`, given that most solutions will have only one project to start with.

It is done for few reasons:
- Familiarity with other non-uno projects, where projects are generally authored this way
- Smaller string to define targets
- Easier string lookup
- Less confusing `DotnetVersion` variable that users may not know where the definition is coming from

This change makes it a bit less easy to change the target framework for the whole solution, but it's a generally known process from developers that use find+replace to do so.

### DIFF
--- a/src/Uno.Templates/content/unoapp-uitest/UnoUITestsLibrary.csproj
+++ b/src/Uno.Templates/content/unoapp-uitest/UnoUITestsLibrary.csproj
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <DotNetVersion Condition=" $(DotNetVersion) == '' ">net8.0</DotNetVersion>
-    <TargetFramework>$(DotNetVersion)</TargetFramework>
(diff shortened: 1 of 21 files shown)


src/Uno.Templates/content/unoapp-uitest/UnoUITestsLibrary.csproj
```diff
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <DotNetVersion Condition=" $(DotNetVersion) == '' ">net8.0</DotNetVersion>
-    <TargetFramework>$(DotNetVersion)</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
dansiegel, line 4: we should never have the actual netX.0 version outside of template.json

dansiegel (approved):

PR #611, opened 2024-04-03
 fixing invalid reference for MauiVersion

dansiegel:

## PR Type

- Bugfix

Versions are located in the Props file where the MauiVersion has not been set and we specify a version that isn't actually available on nuget.org

## What is the new behavior?

The versions are now in the targets and we've updated the MauiVersion to one that actually exists on NuGet.org

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -807,7 +807,7 @@
         "cases": [
           {
             "condition": "(tfm == 'net8.0')",
-            "value": "8.0.5"
+            "value": "8.0.7"
           }
         ]
       }
--- a/src/Uno.Templates/content/unoapp/Directory.Build.props
+++ b/src/Uno.Templates/content/unoapp/Directory.Build.props
@@ -1,4 +1,4 @@
-<Project>
+﻿<Project>
   <PropertyGroup>
     <DotNetVersion>$baseTargetFramework$</DotNetVersion>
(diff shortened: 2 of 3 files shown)


jeromelaban (approved):

src/Uno.Templates/content/unoapp/.template.config/template.json
```diff
@@ -807,7 +807,7 @@
         "cases": [
           {
             "condition": "(tfm == 'net8.0')",
-            "value": "8.0.5"
+            "value": "8.0.7"
```
agneszitte, line 810: @dansiegel, @jeromelaban can we use latest stable **8.0.14** instead for the MauiVersion?

For example with `CommunityToolkit.Maui` latest stable **8.0.0** if following 
https://platform.uno/docs/articles/external/uno.extensions/doc/Learn/Maui/ThirdParty-MauiCommunityToolkit.html

```
Error: Detected package downgrade: Microsoft.Maui.Controls from 8.0.14 to 8.0.7. Reference the package directly from the project to select a different version. 
error NU1605:  MauiEmbeddingApp.Windows -> MauiEmbeddingApp -> MauiEmbeddingApp.MauiControls -> CommunityToolkit.Maui 8.0.0 -> Microsoft.Maui.Controls (>= 8.0.14) 
error NU1605:  MauiEmbeddingApp.Windows -> Microsoft.Maui.Controls (>= 8.0.7)
```

dansiegel: This is a fallback. It should be using the version provided by the Maui workload.

PR #622, opened 2024-04-05
Allow hiding the global json in the library template

dansiegel:

- closes unoplatform/uno-private#424

## PR Type

- Feature

`global.json` is always included in the Library template

## What is the new behavior?

`global.json` is included by default, but there is a new template option that allows you to exclude the `global.json` so that you can easily add the library to an existing solution and use the `global.json` from your existing project.


dansiegel:
@mergifyio backport release/stable/5.2

agneszitte (approved):

Issue #654, opened 2024-04-23
`TargetFrameworks` should be better as a single line


Youssef1313:
![image](https://github.com/unoplatform/uno.templates/assets/31348972/d3791071-c539-4a70-bb5c-e7179221a425)

**If** someone decided to do `<CompilerVisibleProperty Include="TargetFrameworks" />` for some reason, the multi-line will break Roslyn. The way compiler-visible properties are passed to the compiler is by generating an .editorconfig file. But Roslyn will not handle this case properly (known limitation).

**NOTE:** I know it's not common at all that someone makes `TargetFrameworks` a compiler-visible property (usually people would care about `TargetFramework`). But if someone decided to do it. It will break badly.

jeromelaban:
I agree. It's causing more trouble than the readability it improves.

eriklimakc:
Does this mean updating all the samples migrated and in migration process to use single line tfms?

cc @agneszitte @jeromelaban

dansiegel:
@eriklimakc yes it would be good to update the samples to use a single line for the tfms

PR #655, opened 2024-04-24
fix: Move TFMs to single line

jeromelaban:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno.templates/issues/654


Youssef1313 (approved):

dansiegel (changes requested):
This won't work as when the Template Engine encounters `<!--#if (useAndroid)-->` it would remove the entire line

PR #661, opened 2024-04-28
fix: Make windows10.0.19041 last in TargetFrameworks

Youssef1313:
When making the project, if the mobile platforms are not selected, and only desktop, wasm, and WinUI are chosen, then windows10.0.19041 becomes first which then causes an UNOB0012 warning

## PR Type

## PR Checklist

## Other information


Youssef1313:
It's desktop that should actually be last.

Youssef1313:
Okay it's whatever TFM I add first, I get a warning says that it shouldn't be first.

cc @jeromelaban @dansiegel

PR #678, opened 2024-05-08
Templates should be ignored by VS to avoid duplication

### LINKED ISSUE #677: [Bug] Exclude unoapp-uitest, unoapp-uwp, unolib templates from being shown in VS
### Current behavior

These templates are shown in VS when installed from CLI which will cause unneccessary duplication with the templates that come from VS extension

### Expected behavior

All CLI templates are excluded from VS.

### Affected platforms

_No response_

### Template Host

Visual Studio

### Host OS

Windows

### Template Options

Template options are default.

### Template Version

```shell
5.2.31 version of templates is used
```

### Anything else we need to know?

_No response_

dansiegel:

- closes #677 

## PR Type

- Bugfix

Only the unoapp and mauilib templates are excluded from Visual Studio which may result in the other templates from being shown twice in Visual Studio.

## What is the new behavior?

All templates are now excluded from Visual Studio so that only the templates from the VSIX will be displayed.


dansiegel:
@mergifyio backport release/stable/5.2

PR #682, opened 2024-05-08
Adding Rider runtime configurations

### LINKED ISSUE #676: Consider adding .run configurations for unoapp template
## What would you like to be added:
Currently if you will create an unoapp and try to run it with Rider you would probably get an error because by default for each configuration it would select in my case net8.0-ios which wouldn't work with the majority of configurations.
To let Rider correctly match project configuration with the TFM we could ship .run configurations so using unoapp template with Rider would be much easier. Sadly there is no such configuration for VS or at least i can't find any ;(

Here [is](https://www.jetbrains.com/help/rider/Run_Debug_Configuration.html#share-configurations) the documentation about .run configurations.

dansiegel:

- closes #676 

## PR Type

- Feature

We don't provide Rider .run profiles

## What is the new behavior?

We have Rider .run profiles

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/template.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/template.json
@@ -1797,43 +1797,52 @@
         {
           "condition": "(platforms != android)",
           "exclude": [
-            "MyExtensionsApp.1/Platforms/Android/**/*"
+            "MyExtensionsApp.1/Platforms/Android/**/*",
+            ".run/MyExtensionsApp.1 (Android).run.xml"
           ]
         },
         {
           "condition": "(platforms != ios)",
           "exclude": [
-            "MyExtensionsApp.1/Platforms/iOS/**/*"
+            "MyExtensionsApp.1/Platforms/iOS/**/*",
+            ".run/MyExtensionsApp.1 (iOS).run.xml"
           ]
         },
         {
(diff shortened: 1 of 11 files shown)


jeromelaban (approved):

Takoooooo:
@dansiegel Great that you managed to find how to include other configurations besides the ones I provided but sadly IOS, Android and MacOS configurations are incorrect. If you will try to run them from Rider you will see the error appearing. I assume IOS, Android and MacOS configs are not really needed because for me Rider picks the correct TFM's for these targets correctly without any configurations(sorry i noticed it a bit late). Though if you think they are needed i have generated ones that work for me [Uno2.run.zip](https://github.com/unoplatform/uno.templates/files/15279582/Uno2.run.zip)
Also it turns out you can put all configs into the single file

src/Uno.Templates/content/unoapp/.template.config/template.json
```diff
@@ -664,6 +664,13 @@
       "defaultValue": "true",
       "description": "Sets up Visual Studio Code Debugging for WebAssembly and Skia Desktop"
     },
+    "rider": {
```
jeromelaban, line 667: Let's make this new option turned on by default, we'll disable it by default in VS's wizard. Let's also add a readme file to explain what those files are.

dansiegel:
replaced by #694

PR #694, opened 2024-05-23
Add Rider runtime configurations v2

Takoooooo:
I've decided to not include the .run configs for iOS,Android and MacCatalyst since they are included by default for me. If that's wouldn't be true for some of our users we can include them later anyway

GitHub Issue (If applicable): closes https://github.com/unoplatform/uno.templates/issues/676
based on https://github.com/unoplatform/uno.templates/pull/682

## PR Type
- Feature

We don't provide Rider .run profiles

## What is the new behavior?

We have Rider .run profiles


Takoooooo:
Btw,@jeromelaban what did you mean by "Let's also add a readme file to explain what those files are"? Not sure where this file should be located and how should it work

Takoooooo:
@dansiegel I've tested all configs on my machine, but it would be nice if you could test and review it on your machine too

Takoooooo:
Closing, because Rider will support that soon-ish

https://youtrack.jetbrains.com/issue/RIDER-111470/Uno-Platform-5.2-new-targets-frameworks-are-not-supported#focus=Comments-27-9887747.0-0

Takoooooo:
Keeping the PR until we'll actually implement that feature on our side

carldebilly:
❤️❤️❤️

PR #731, opened 2024-06-06
ci: Disable automerge

jeromelaban:
Disable automerge for now, until the process becomes stable enough.


dansiegel (approved):

dansiegel:
We should probably also update the build to ignore changes to the github workflows

PR #756, opened 2024-06-17
ci: Updater fixes for release branches

dansiegel:

- n/a

## PR Type

- Project automation

Uno Sdk Updater is really tooled to run for dev only
Uno Sdk Updater runs once per hour

## What is the new behavior?

Uno Sdk Updater now uses a matrix to potentially create one or more PRs for dev and stable releases
Uno Sdk Updater now runs once every 6 hours. This will reduce the load on CI agents as some update PRs have triggered new builds due to a 2nd update just before the initial set of CI validations finished.


.github/workflows/uno-updater.yml
```diff
@@ -20,12 +19,29 @@ on:
 jobs:
   manifest-update:
     runs-on: windows-latest
+    strategy:
+      matrix:
+        branch:
+          - main
+          - release/stable/5.3
```
jeromelaban, line 26: Let's use that instead, I thought it was used by triggers, but it's not.

```suggestion
          - release/stable/*
```

jeromelaban: Scratch that, wrong line.

.github/workflows/uno-updater.yml
```diff
@@ -2,11 +2,10 @@ name: Update Uno Sdk
 
 on:
   schedule:
-    - cron: '0 * * * *'
+    - cron: '0 */6 * * *'
   pull_request:
     branches:
       - main
-      - release/stable/5.3
```
jeromelaban, line 9: I think we should still be executing it for release branches, with `release/stable/*`.

dansiegel: Just because we have a backport to a release branch doesn't mean the SDK updater needs to get triggered.

jeromelaban (approved):

jeromelaban:
@mergifyio backport release/stable/5.3

PR #922, opened 2024-08-21
Aligning template.json & templatewizard.json

### LINKED ISSUE #911: [Bug] Template generated with preset "recommended" doesen't match the TemplateWizard.json values and Live Wizard
### Current behavior

I assume that both CLI templates and Live Wizard should be synced based on what we have declared in the TemplateWizard.json for the recommended preset
image

1) Tests are disabled on the GUI for the recommended template in Live Wizard
MicrosoftTeams-image
But when i generate the app with simply -preset "recommended" i defenitely see Unit and UI tests being added
MicrosoftTeams-image (1)
Related place in the code - https://github.com/unoplatform/uno.templates/blob/main/src/Uno.Templates/content/unoapp/.template.config/template.json#L348-L349

2) Theme Service is enabled on the UI in Live Wizard and disabled in the generated template by the recommended preset
image
Here is the UnoFeatures section for the template generated from CLI with recommended preset and it doesen't contain the ThemeService feature
image
I think we simply don't have generated property which will change the value based on the preset so the value of themeService property is used which is false by default.
https://github.com/unoplatform/uno.templates/blob/main/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MyExtensionsApp.1.csproj#L102-L104
https://github.com/unoplatform/uno.templates/blob/main/src/Uno.Templates/content/unoapp/.template.config/template.json#L578-L584

3) LoggingSerilog is added to the features of the generated template while i assume it shouldn't
image
image
Related place in code -
Here we are expecting the serilog to be the default value for recommended preset
https://github.com/unoplatform/uno.templates/blob/main/src/Uno.Templates/content/unoapp/.template.config/template.json#L937-L954

### Template Host

dotnet new

### Host OS

Windows

### Template Options

dotnet new unoapp -o ewwed -preset "recommended" -pub "ewwed"

### Template Version

```shell
5.3.96
```

dansiegel:

- closes #911 

## PR Type

- Bugfix

The Template.json and TemplateWizard.json are not aligned on the definition of Blank and Recommended.

## What is the new behavior?

Both should be aligned now with what options are set for Blank vs Recommended.


dansiegel:
@mergifyio backport release/stable/5.3

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
```diff
@@ -68,7 +68,7 @@
 					"architecture": "none",
 					"authentication": "none",
 					"configuration": "false",
-					"cultures": "",
+					"cultures": "es|fr|pt-BR",
```
nickrandolph, line 71: Blank shouldn't specify cultures

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
```diff
@@ -78,6 +78,7 @@
 					"mediaElement": "false",
 					"navigation": "blank",
 					"platforms": "android|ios|maccatalyst|wasm|windows|desktop",
+          "rider": "true",
```
nickrandolph, line 81: What does this parameter do? Similar to IsVsix, if this is specific to Rider, it shoudln't be in here

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
```diff
@@ -126,7 +128,7 @@
 					"mauiEmbedding": "false",
 					"wasmMultiThreading": "false",
 					"dspGenerator": "true",
-					"themeService": "true",
+					"themeService": "false",
```
nickrandolph, line 131: This should default to true in the wizard experience for recommended, which is why it should be true

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
```diff
@@ -111,11 +112,12 @@
 					"http": "true",
 					"isVsix": "true",
 					"localization": "true",
-					"logging": "default",
+					"logging": "serilog",
```
nickrandolph, line 115: I'm not sure about setting serilog to be the default in the recommended template. Perhaps we should back off the template to using default instead of changing the wizard behavior

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
dansiegel: I don't disagree with you... however, as I recall it's hard coded that way on the CLI because we couldn't get it to work where one way we had one behavior and the other we have the other behavior. Thus it's more important to ensure that the behavior is consistent.

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
dansiegel: So this actually defaults to true on the CLI. It adds a `.rider` folder with some configurations that allow you to run the project in Rider... similar to the VS Code stuff we have.

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
dansiegel: This was always the default in Recommended, the TemplateWizard went rogue.

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
dansiegel: This was always the default in Recommended, the TemplateWizard went rogue.

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
nickrandolph: Agreed. VSIX needs to set isVsix=true and rider=false as overrides to this file

nickrandolph (approved):

PR #1059, opened 2024-10-26
fix: CSWinRT-related compilation issues

Takoooooo:
This one could be reverted when this issue is fixed -> https://github.com/microsoft/CsWinRT/issues/1809
Decided to use CsWinRTAotOptimizerEnabled instead of AllowUnsafeBlocks since it seems that there are other related issues to WinRT.SourceGenerator

### DIFF
--- a/src/Uno.Templates/content/unoapp/Directory.Build.props
+++ b/src/Uno.Templates/content/unoapp/Directory.Build.props
@@ -3,6 +3,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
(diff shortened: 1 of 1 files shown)


src/Uno.Templates/content/unoapp/Directory.Build.props
```diff
@@ -3,6 +3,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CsWinRTAotOptimizerEnabled>false</CsWinRTAotOptimizerEnabled>
```
jeromelaban, line 6: Thanks. We won't this this in the templates directly at this time, we'll adjust that in the SDK.

PR #1184, opened 2025-01-23
fix(vs): Adjust tfm order for vs17.12

jeromelaban:
This change will make 17.11 show a warning message for UNOB0015.

GitHub Issue (If applicable): closes https://github.com/unoplatform/private/issues/599

PR #1196, opened 2025-01-27
Revert "fix(vs): Adjust tfm order for vs17.12"

jeromelaban:
Reverts unoplatform/uno.templates#1184

Placing desktop first breaks the WinAppSDK debugging.


agneszitte:
@jeromelaban this PR would need to be backported to 5.6 I imagine?

PR #1203, opened 2025-01-29
fix: Use "new" android bridge by default

jeromelaban:
Follows the recommendation from https://github.com/dotnet/runtime/issues/106410#issuecomment-2291508507


Syed-RI:
Does it "Actually" fixes the issue? were you able to confirm it?

PR #1298, opened 2025-03-28
use RelayCommandAttribute in uno templates when using the MVVM pattern

corvinsz:
GitHub Issue (If applicable): -

## PR Type

Currently the ICommand properties are not source generated.

## What is the new behavior?
The ICommand properties are being source generated.

## PR Checklist

## Other information

I recently created an uno project using the templates you guys provide via https://new.platform.uno/.
I noticed when selecting the `MVVM` pattern, it does not utilize the [RelayCommandAttribute](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/relaycommand) from the [CommunityToolkit.Mvvm](https://www.nuget.org/packages/CommunityToolkit.Mvvm).
Unless there are uno specific limitations or guidelines I'm not aware of, I don't see a reason to not use the source generated property.

I'm new to uno, so bear with me if what I'm proposing is total nonsense.

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginPage.xaml
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginPage.xaml
@@ -33,7 +33,7 @@
             HorizontalAlignment="Stretch" />
 <!--#endif-->
       <Button Content="Login"
-          Command="{Binding Login}"
+          Command="{Binding DoLoginCommand}"
           HorizontalAlignment="Stretch" />
     </StackPanel>
   </Grid>
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginPage.xaml.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Presentation/LoginPage.xaml.cs
@@ -47,7 +47,7 @@ public sealed partial class LoginPage : Page
                             new Button()
                                 .Content("Login")
                                 .HorizontalAlignment(HorizontalAlignment.Stretch)
-                                .Command(() => vm.Login)))));
+                                .Command(() => vm.DoLoginCommand)))));
 #else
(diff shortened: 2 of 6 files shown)


DevTKSS:
@corvinsz did you also get the observable Properties working? I just could decide between compiler telling me I would try to use reflection and should use aot compartble definition or it wanted me to use the .net 9 fields... Like I tryed every community Toolkit described scenario for them and nothing was not giving me a warning.
Why did you change the names of them? I mean the Login to "DoLogin" and so on... That would not be anything I would expect...

corvinsz:
@DevTKSS I assume you mean using the new partial properties which got released with .NET 9? I have not tried that, no.
>Why did you change the names of them? I mean the Login to "DoLogin" and so on... That would not be anything I would expect...

I changed the command-bindings from `Login` to `DoLogin` because that's how the source generator works. The method is called `DoLogin` and applying the CommunityToolkit.MVVM attribute `[RelayCommand]` to it, generates an ICommand (or an async version of that) which is called `<Method name>Command`.

DevTKSS:
Oh that's interesting! Did not recognize this detail so far on the command and relay commands...
For the property attribute, I am struggling how this would be used in uno mvvm app but would not result in a complaining Compiler.
Following the community Toolkit mvvm gallery guidance, I tryed out all of them and the ones intellisence did tell but might be that I did it wrong so I did not find a way to have it not complaining... Last time I created a mvvm uno app this attribute did miss on the property. If you do know how to add it correctly, I at least would like this already implemented in the template 🤔 would make sense somehow if the rest of the app is using mvvm the properties does also

corvinsz:
@DevTKSS I'm not super familiar with uno itself, but I just spun up a uno project and was able to successfully data-bind a partial property. In order for this to work you have to set the `<LangVersion>preview</LangVersion>` in your project. If you need help with that you can reach out to me on discord (name: "qlay"), I don't think this PR is the correct place to discuss problems like that. 😄

Issue #619, opened 2024-04-05
Enable build with warnings as errors for template validations


jeromelaban:
This will prevent issues for breaking changes or deprecations to be avoided in the default templates.

PR #1499, opened 2025-05-19
chore: Enable TreatWarningsAsErrors

Takoooooo:
Fixes https://github.com/unoplatform/uno.templates/issues/619

### DIFF
--- a/.github/actions/ci/run-tests/action.yml
+++ b/.github/actions/ci/run-tests/action.yml
@@ -344,7 +344,8 @@ runs:
             "-p:WasmShellILLinkerEnabled=false",
             "-p:UseInterpreter=true",
             "-p:_IsDedupEnabled=false",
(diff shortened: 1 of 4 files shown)


Takoooooo:
These -
https://github.com/unoplatform/uno.templates/actions/runs/15109544379/job/42465733482?pr=1499#step:3:1537
https://github.com/unoplatform/uno.templates/actions/runs/15109544379/job/42465733906?pr=1499#step:3:1117
Should get fixed after 
https://github.com/unoplatform/uno.extensions/pull/2773
Apparently we had a not needed reference to Uno.Extensions.Navigation coming from Uno.Extensions.Authentication which caused ForceBindingsUpdateGenerator to be fired and throw an exception since ForceBindingsUpdateGenerator needed more packages to work correctly

jeromelaban (approved):

PR #1548, opened 2025-06-04
chore: Update SkiaSharp to stable version

nickrandolph:
Updating SkiaSharp to stable release


MartinZikmund:
The core PR was merged, so this can be finalized for 6.1

nickrandolph:
This is no longer relevant as main is on stable version already now

PR #1552, opened 2025-06-04
chore: update maui parts of app templates

kazo0:


src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MainPage.xaml
```diff
@@ -8,14 +8,10 @@
...
     <Grid$toolkitSafeArea$>
-      <StackPanel
-        HorizontalAlignment="Center"
-        VerticalAlignment="Center">
-        <TextBlock AutomationProperties.AutomationId="HelloTextBlock"
-          Text="Hello Uno Platform!"
-          HorizontalAlignment="Center" />
 <!--#if (mauiEmbedding)-->
 <!--#if (useNonMauiPlatforms)-->
+      <Grid HorizontalAlignment="Center"
```
nickrandolph, line 13: Why are we changing the root panel from StackPanel to Grid - the difficulty with a Grid is that if the user adds controls (eg in HD) they will stack on top of each other by default. Hence why the template has a StackPanel

kazo0: It's due to a limitation with embedding native views in a skia-rendered app. 

cc @ramezgerges for explanation

nickrandolph: Can we limit this to only change to Grid when embedding maui?

ramezgerges: Responding to Steve's cc: The problem is that on most platforms, the hosted native element will expand to fit whatever space is given to it during measure, so if you give it infinite bounds, things will go bad, hence the change to a Grid with a Center alignment.

nickrandolph: My preference would be to keep stackpanel and just put a height/maxheight on the maui embedding control - this is sample code anyhow, so I don't see an issue with this.

PR #1615, opened 2025-06-25
chore: enable SkiaRenderer by default for unolib

### LINKED ISSUE #1614: Enable skiarenderer for libraries by default
We are currently not enabling SkiaRenderer as an UnoFeature by default, which can cause errors when building with an application head that is using SkiaRenderer. As seen here: https://github.com/unoplatform/uno/discussions/20850

The `unolib` template should be generating a project, conditionally on the renderer option, to use the SkiaRenderer UnoFeature. Similar to how we are doing it for the `unoapp` template: https://github.com/unoplatform/uno.templates/blob/bd9c2f7e144ecc52a6711fd8211675c2321a6760/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/MyExtensionsApp.1.csproj#L113-L115

Additionally, we should investigate how this was not caught by any automated tests and see if we are missing any test cases here

Kunal22shah:
GitHub Issue (If applicable): closes #1614

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/.github/actions/ci/generate-test-matrix/action.yml
+++ b/.github/actions/ci/generate-test-matrix/action.yml
@@ -45,8 +45,10 @@ runs:
 
             "mauilib_MacOS;macOS;unomauilib;"
             "unolib_MacOS;macOS;unolib;"
+            "unolibNative_MacOS;macOS;unolib;--renderer native"
             "mauilib_Windows;Windows;unomauilib;"
             "unolib_Windows;Windows;unolib;"
+            "unolibNative_Windows;Windows;unolib;--renderer native"
 
             # Disabling this test for now as it was specific to net7
             # https://github.com/unoplatform/uno.templates/issues/22
--- a/src/Uno.Templates/content/unolib/.template.config/template.json
+++ b/src/Uno.Templates/content/unolib/.template.config/template.json
@@ -60,7 +60,23 @@
       "type": "parameter",
       "datatype": "bool",
       "defaultValue": "true"
-    }
+    },
+    "renderer": {
+    "type": "parameter",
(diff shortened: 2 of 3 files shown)


src/Uno.Templates/content/unolib/.template.config/template.json
```diff
@@ -60,7 +60,23 @@
       "type": "parameter",
       "datatype": "bool",
       "defaultValue": "true"
-    }
+    },
+    "renderer": {
```
jeromelaban, line 64: Let's add a test in the matrix for these options.

.github/actions/ci/generate-test-matrix/action.yml
```diff
@@ -45,8 +45,10 @@ runs:
 
             "mauilib_MacOS;macOS;unomauilib;"
             "unolib_MacOS;macOS;unolib;"
+            "unolibNative_MacOS;macOS;unolib;--renderer native"
             "mauilib_Windows;Windows;unomauilib;"
             "unolib_Windows;Windows;unolib;"
```
kazo0, line 50: I'm curious why this existing test wasn't failing prior to this change when attempting to be built for WASM?

@jeromelaban it should have been failing, no?

Or is this specifically just creating an unolib project but not incorporating it as part of another Uno application project?

Do we have a way to easily generate that kind of application? like an unoapp with an unolib inside?

jeromelaban: This is only a "new lib", not a new project with a library dependency. It's a scenario that is not covered by the test matrix yet.

Issue #1606, opened 2025-06-21
[Docs][Testing][!Assertation] Add information about FluentAssertions Package is Included in all Testing Projects!


DevTKSS:
## What would you like clarification on:

Please add information to the Testing Docs, that Uno Testing Projects, coming from the Template Wizard / dotnet new Templates, are including FluentAssertion Package for Assertation.

## Concern?

- [x] Usage in industry
- [ ] Clarification of capabilities
- [ ] Getting started with Uno
- [x] Developing with Uno
- [ ] Contributing to the Uno project
- [x] Publishing your application
- [ ] Support
- [ ] Other (please specify):

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [x] Testing

## Anything else we need to know?

**The License for any version up from 8.0 is required even if you don't the purpose or do make profit using it!**

https://github.com/fluentassertions/fluentassertions/blob/7928d551ec2615a98b19d51161e374641ecd4fb0/LICENSE#L21C103-L21C182

Especially for Companys, they should know about this, as the fee has to get payed **PER DEVELOPER**:
https://xceed.com/products/unit-testing/fluent-assertions/

DevTKSS:
@jeromelaban would you move this issue to the Repo where we have the appropriate docs living, so it gets seen and that content added to the docs? seems like they are not in this repo

PR #1660, opened 2025-07-18
chore(FluentAssertions): Update provided Version to latest free and pin it there

### LINKED ISSUE #1625: Problem prevention Request: Update and Pin FluentAssertions Package to latest free version for provided Templates
## Issued Code

https://github.com/unoplatform/uno.templates/blob/14b349b87cb182616084e9a30182c977fae38baa/src/Directory.Build.targets#L10

## What should get changed

Reffering to #1606 , I would like to request the test project provided Version of FluentAssertions to:

- *(optional)* Update to `7.2.0` which is the latest **FREE & OPEN SOURCE** Version
- **Pin the provided version** `<PackageVersion Include="FluentAssertions" Version="[7.0.2]"/>`
- Add a comment line thats informing the user about, when considering to update this, that the versions above requires paid license

## Why is that needed

nuget package manager in vs 2022 for example suggests to update the Package as there is a higher version included. If the user might do this with `dotnet package update ...` CLI command, maybe the tool will not even prompt him like the Package man does kind of inform us that they are not relyable for any License the used or updated Packages might require for legal use

## Reference Sourced from:

https://youtu.be/ZFc6jcaM6Ms?si=1MrgTx1rKkOBa0l_

## Any other Informations

As I can understand, that the reffered Issue will be a way higher Effort, maybe a Version pin could be a workaround until the refered issue got evaluated for bigger effort change

DevTKSS:
GitHub Issue (If applicable): closes #1625 

## PR Type

- Dependency update for the provided Templates
- User/Consumer Expirience improvement
- Build or CI related changes

- Package Version is lower than needed
- Using the provided github actions dependabot routine, the user of the template could get into licensing trouble, when not checking that version for now be payed license

## What is the new behavior?

- Updated the Version to the latest free (Apache2.0) version
- pinning the version of this particular Package

   By pinning that version, the user still has the active choice, to remove the Brackets, if he might want to pay for newer versions, but those who may don't want to, but using GitHub actions and dependabot, will not run into licensing problems by accident. This results also in a better user experience

## PR Checklist

## Other information

Maybe there should be added a note in the documentation thats informing the user about this, but kind of all testing documentation are in contribution section, which makes it not clearly visible for non-contributor users
Maybe that should get added in the templating docs, but those are not on this Repository => can not be included in this PR by that. maybe move the #1606 issue to the appropriate repo, which holds that docs and resolve it there?

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -7,7 +7,7 @@
     <MicrosoftTestSdkVersion Condition="'$(MicrosoftTestSdkVersion)' == ''">17.9.0</MicrosoftTestSdkVersion>
     <NUnitVersion Condition="'$(NUnitVersion)' == ''">4.1.0</NUnitVersion>
     <NUnit3TestAdapterVersion Condition="'$(NUnit3TestAdapterVersion)' == ''">4.5.0</NUnit3TestAdapterVersion>
-    <FluentAssertionsVersion Condition="'$(FluentAssertionsVersion)' == ''">6.12.0</FluentAssertionsVersion>
+    <FluentAssertionsVersion Condition="'$(FluentAssertionsVersion)' == ''">[7.2.0]</FluentAssertionsVersion>
     <CoverletCollectorVersion Condition="'$(CoverletCollectorVersion)' == ''">6.0.2</CoverletCollectorVersion>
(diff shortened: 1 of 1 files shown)


DevTKSS:
@jeromelaban what about this?

DevTKSS:
Not sure what is causing this to fail🤔

https://github.com/unoplatform/uno.templates/actions/runs/16380313603/job/46291634241#step:2:1

PR #1673, opened 2025-07-24
chore: Remove link to solved issue

DevTKSS:

## PR Type

if the user includes the Mobile target, he will get this comment added in his template, advising to uncomment/comment out the Packaged WinAppSdk target, but this seems (following this link) to be fixed 🎉  lets remove this? 🙃 

## What is the new behavior?

lets remove this? 🙃 

## PR Checklist

## Other information

While this particular issue is fixed, could the team quick check if there are other issues open (Backlog I think you call it?) that would still require us users to do anything on this profile and we might should have a comment to this there, but with some link update?

### DIFF
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Properties/launchSettings.json
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Properties/launchSettings.json
@@ -31,9 +31,6 @@
 //#endif
 //#if (useWinAppSdk)
 //#if (!useMsalAuthentication)
-//#if (useMobile)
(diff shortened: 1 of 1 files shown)


src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Properties/launchSettings.json
```diff
@@ -31,9 +31,6 @@
 //#endif
 //#if (useWinAppSdk)
 //#if (!useMsalAuthentication)
-//#if (useMobile)
-    // Note: In order to select this profile, you'll need to comment the `Packaged` profile below until this is fixed: https://aka.platform.uno/wasdk-maui-debug-profile-issue
```
kazo0, line 35: Just want to double check and confirm with @jeromelaban that this comment is indeed no longer needed

PR #2084, opened 2026-05-01
chore: make simple theme default and change icon

### LINKED ISSUE #2075: Add Simple theme option
## What would you like to be added:

Add a new "Simple" theme choice to the `appTheme` parameter in the Uno app template. This would be a lightweight theme with no design system dependency, appearing as the first option before Material, Fluent, and Cupertino.

Specifically, the following changes are needed:

- In `template.json`: add a `"simple"` choice to the `appTheme` symbol with displayName "Simple"
- In `TemplateWizard.json`: add `"simple"` to the Theme section's `Choices` array and add an `appTheme.simple` metadata entry with a `Sequence` value lower than Material (e.g., `0`) so it appears first in the wizard UI

## Why is this needed:

Uno Studio's project creation wizard needs a "Simple" theme option for users who want a minimal starting point without a full design system (Material, Fluent, or Cupertino). The icons and wizard UI support are being added on the Uno Studio side, but the template definitions need to include the "simple" choice for it to appear in the wizard.

The Uno Studio side will already have the icon assets (`Theme.Simple.svg`) in all four variants (light, dark, disabled, dark-disabled) ready to go. The studio wizard reads theme choices from `template.json` and display metadata from `TemplateWizard.json`, so both files need the new entry for the option to surface.

rajamatt:
GitHub Issue (If applicable): closes #2075 

## PR Type

- Refactoring

The Simple theme (`simpletheme`) appears third in the wizard theme list (Sequence 3, after Material and Fluent) and uses the Cupertino icon as a placeholder.

## What is the new behavior?

- The Simple theme now appears first in the wizard theme list (Sequence 1)
- The Simple theme uses its own icon (`Theme.Simple.svg`)
- Theme sequence is now: Simple (1), Material (2), Fluent (3), Cupertino (4)


nickrandolph (approved):

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
```diff
@@ -360,16 +360,16 @@
...
 			"Icon": "/Assets/Theme.Fluent.svg"
 		},
 		"appTheme.material": {
-			"Sequence": 1,
+			"Sequence": 2,
 			"Icon": "/Assets/Theme.Material.svg"
 		},
-		"appTheme.simpletheme": {
-			"Sequence": 3,
-			"Icon": "/Assets/Theme.Cupertino.svg"
```
ajpinedam, line 372: If the Cupertino.svg asset is not longer being used, should it be removed?

ajpinedam (approved):

src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
rajamatt: Good point, but this was just a placeholder for simple theme. The Cupertino theme is still there with the `Assets/Theme.Cupertino.svg` (line 360 of this file)

PR #2090, opened 2026-05-04
chore(android): Bump AppCompat to 1.7.1.2 and minSdk to 23

MartinZikmund:
## Summary
- Bumps `Xamarin.AndroidX.AppCompat` to `1.7.1.2` for both the default and `net10.0` overrides in the templates' Sdk `packages.json`.
- Raises the Android `SupportedOSPlatformVersion` to `23.0` in the `MyExtensionsApp.1.MauiControls` template csproj.

Aligns the templates with the AppCompat 1.7.1.2 rollout in the main Uno Platform repo. AppCompat 1.7.x requires API 23 as the minimum Android level, so projects scaffolded from these templates need to match.

## Test plan
- [ ] Verify a fresh `unoapp` template scaffold builds for `net10.0-android`
- [ ] Verify the `MauiControls` variant compiles with the new `SupportedOSPlatformVersion`

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/Uno.Sdk/packages.json
+++ b/src/Uno.Sdk/packages.json
@@ -240,12 +240,12 @@
   },
   {
     "group": "AndroidXActivity",
-    "version": "1.10.1.2",
+    "version": "1.12.2.1",
     "packages": [
       "Xamarin.AndroidX.Activity"
     ],
     "versionOverride": {
-      "net10.0": "1.10.1.3"
+      "net10.0": "1.12.2.1"
     }
   },
   {
--- a/src/Uno.Sdk/packages.json
+++ b/src/Uno.Sdk/packages.json
@@ -220,12 +220,12 @@
   },
   {
     "group": "AndroidXAppCompat",
-    "version": "1.7.0.7",
+    "version": "1.7.1.2",
     "packages": [
       "Xamarin.AndroidX.AppCompat"
     ],
     "versionOverride": {
-      "net10.0": "1.7.1.1"
+      "net10.0": "1.7.1.2"
     }
   },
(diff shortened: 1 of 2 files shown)


MartinZikmund:
Closing as superseded — **with one piece that is not carried over, flagged below.**

The binding bumps are covered by #2246, which goes further on both groups:

| group | here | #2246 |
|---|---|---|
| `AndroidXAppCompat` (net10.0) | 1.7.1.2 | **1.8.0** |
| `AndroidXActivity` (net10.0) | 1.12.2.1 | **1.13.0.1** |

### ⚠️ Not carried over: the `minSdk` change

The second half of this PR — `SupportedOSPlatformVersion` 21.0 → 23.0 in `MyExtensionsApp.1.MauiControls.csproj` — is **not** in #2246 and is not on `main`. It is deliberately left here rather than folded in, because the proposed value looks stale:

- `MyExtensionsApp.1.MauiControls` is a plain `Microsoft.NET.Sdk` MAUI class library, so it does **not** inherit the Uno.Sdk's Android floor and has to state its own.
- The Uno.Sdk already defaults Android to **24.0** (`Uno.Common.Android.targets`, and 26.0 with the `androidwear` feature). So every other head in the same generated app is at 24, and this project at 21 is the outlier — 23 would still be below the app's own floor.

So the open question isn't "21 or 23" but whether this project should simply align on **24.0** (or drop the override). That's a separate call from an AndroidX version sweep, which is why it isn't being smuggled into #2246. Worth its own PR.

PR #2093, opened 2026-05-05
feat: Enhance navigation initialization for shell support in App class

kazo0:
This pull request updates the app launch logic in `App.recommended.cs` to improve how navigation is initialized, especially when using the shell template and authentication. The changes streamline navigation setup and ensure that the correct initial route or view model is loaded based on authentication status.

Navigation initialization improvements:

* When using the shell template without authentication, navigation is now initialized via `MainWindow.InitializeNavigationAsync`, specifying the initial route as "Main" for a clearer and more consistent startup experience.
* When using the shell template with authentication, navigation is initialized with a new `initialNavigate` delegate. This delegate checks authentication status and navigates to either the main or login view model accordingly, ensuring users are directed to the appropriate screen on launch.


kazo0:
@copilot+claude-sonnet-4.6 apply changes based on the comments in [this thread](https://github.com/unoplatform/uno.templates/pull/2093#pullrequestreview-4225263402)

hiadamhere (approved):

PR #2125, opened 2026-05-27
feat: add dotnet new item templates (Page, Window, ContentDialog, …)

MartinZikmund:
## What

Adds `dotnet new` **item templates** to `Uno.Templates` — the CLI / Rider / VS Code "Add New Item" story that was previously missing (VS-only `.vstemplate` items live in the separate uno.studio repo).

### Item templates
- `uno-page`, `uno-window`, `uno-usercontrol`, `uno-contentdialog`, `uno-resourcedictionary`, `uno-resourcedictionary-codebehind`, `uno-resw`, `uno-templatedcontrol`
- **`uno-templatedcontrol` ships a working default `Themes/Generic.xaml`** (a real fix vs. the uno.studio item, which shipped none — the control rendered nothing), with an `-I` / `--include-default-style` switch to skip it on re-runs.
- **C# Markup variants** (`-markup csharp`) for page/window/usercontrol/contentdialog.
- **MVVM/MVUX paired pages**: `uno-mvvm-page`, `uno-mvux-page`.

All item templates set `ide.host.json` `unsupportedHosts: [vs]` so Visual Studio keeps using its existing uno.studio items (no duplicate "Add New Item" entries).

### Validation
`build/test-item-templates.ps1` generates `unoapp` for 3 preset/markup combos, adds every item template, and builds the desktop head — **3/3 pass** (including the MVUX `Bindable` ctor and the C# Markup bodies). A dedicated `item-template-smoke-test` CI job runs it.

### Docs
`docs/item-templates.md` + README link.


docs/item-templates.md
```diff
@@ -0,0 +1,29 @@
...
+| `dotnet new uno-resourcedictionary -n MyDictionary` | A `ResourceDictionary` (no code-behind). |
+| `dotnet new uno-resourcedictionary-codebehind -n MyDictionary` | A `ResourceDictionary` with code-behind. |
+| `dotnet new uno-resw -n Resources` | A `.resw` resource file for localized strings. |
+| `dotnet new uno-templatedcontrol -n MyControl` | A lookless `Control` plus a default style in `Themes/Generic.xaml`. Pass `-I false` to skip `Generic.xaml` when it already exists. |
+| `dotnet new uno-mvvm-page -n MyPage` | A `Page` + a CommunityToolkit.Mvvm `ViewModel`. |
+| `dotnet new uno-mvux-page -n MyPage` | A `Page` + an Uno.Extensions.Reactive (MVUX) `Model`. |
+
+Pass `--root-namespace <ns>` (or `-ns <ns>`) when generating outside a project
+to control the generated namespace; inside a project it is inferred from the
+project's `RootNamespace`.
```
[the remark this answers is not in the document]
MartinZikmund: Good catch — the docs were describing behaviour the templates didn't have. Fixed the templates rather than the docs, in 8f9cc8a6 (+ test in 34a1ca20).

All 8 item templates that emit a namespace now bind to the host project and coalesce, the same shape the .NET SDK's own `class` item template uses:

```json
"userNamespace":     { "type": "parameter", "datatype": "string", "defaultValue": "" },
"msbuildNamespace":  { "type": "bind", "binding": "msbuild:RootNamespace" },
"fallbackNamespace": { "type": "generated", "generator": "constant", "parameters": { "value": "UnoApp" } },
"projectNamespace":  { "type": "generated", "generator": "coalesce",
                       "parameters": { "sourceVariableName": "msbuildNamespace", "fallbackVariableName": "fallbackNamespace" } },
"rootNamespace":     { "type": "generated", "generator": "coalesce",
                       "parameters": { "sourceVariableName": "userNamespace", "fallbackVariableName": "projectNamespace" },
                       "replaces": "$rootnamespace$" }
```

Precedence is `--root-namespace` → host project's `RootNamespace` → `UnoApp`. `dotnetcli.host.json` now maps `--root-namespace`/`-ns` onto `userNamespace` so the CLI surface is unchanged.

Verified against a real project rather than by inspection — in an `Acme.Widgets` classlib:

| invocation | namespace / x:Class |
|---|---|
| `dotnet new uno-page -n InferredPage` | `Acme.Widgets` / `Acme.Widgets.InferredPage` |
| `dotnet new uno-page -n ExplicitPage -ns Contoso.Custom` | `Contoso.Custom` / `Contoso.Custom.ExplicitPage` |
| same, outside any project | `UnoApp` / `UnoApp.OrphanPage` |

`build/test-item-templates.ps1` now asserts the inferred and overridden cases — a passing build alone wouldn't catch the binding silently reverting to the fallback. Docs reworded to lead with the inference and note the `UnoApp` fallback.

build/test-item-templates.ps1
```diff
@@ -0,0 +1,102 @@
...
+    $appName = (($Name -split '[^A-Za-z0-9]') | Where-Object { $_ } |
+        ForEach-Object { $_.Substring(0, 1).ToUpper() + $_.Substring(1) }) -join ''
+
+    Push-Location $appDir
+    try {
+        Invoke-Dotnet new unoapp -preset $Preset -presentation $Presentation -markup $Markup -platforms desktop -n $appName -o . --force
+
+        $markupArgs = @("-markup", $Markup)
+
+        Invoke-Dotnet new uno-page -n SampleItemPage @markupArgs
```
morning4coffe-dev, line 68: [P2] The working directory is still $appDir, but unoapp -n $appName -o . creates the application project under $appDir/$appName. These item commands therefore write beside the project, not into its Compile/Page glob. I reproduced this with the actual template engine and a restored nested project: generating from the solution root produced namespace UnoApp and the project's evaluated Compile list stayed empty; generating from the restored project directory correctly inferred its RootNamespace. Enter the generated project directory before adding/asserting/building the items, and assert that the new files are actual project inputs.

.github/workflows/ci.yml
```diff
@@ -219,6 +219,61 @@ jobs:
...
+    - name: 'Install Project Templates'
+      shell: pwsh
+      run: |
+        dotnet new install "${{ github.workspace }}/src/PackageCache/Uno.Templates*.nupkg"
+
+    - name: 'Setup nuget sources'
+      shell: pwsh
+      run: |
+          dotnet nuget add source -n nuget_local ${{ github.workspace }}/src/PackageCache
+          dotnet nuget add source -n uno_dev "https://pkgs.dev.azure.com/uno-platform/1dd81cbd-cb35-41de-a570-b0df3571a196/_packaging/unoplatformdev/nuget/v3/index.json"
```
morning4coffe-dev, line 271: [P2] The repository nuget.config already contains this unoplatformdev URL. Adding it again under uno_dev fails with 'The source specified has already been added' (exit 1), which I reproduced against an isolated copy of the config. Unlike the older setup step that accidentally masks this with a later successful command, this is the new step's last command, so it fails before the smoke test. Keep the genuine local-package source add, but remove or make this inherited-source add idempotent.

morning4coffe-dev (changes requested):
The item templates generate correctly inside a restored project, including inferred/explicit namespaces and the XAML/C# Markup variants I checked. The new validation path has two real problems, however: it adds the items at the generated solution root rather than inside the application project, and the new CI source-setup step ends with a duplicate feed add that fails. Both need correcting before the claimed compile coverage is reliable.

PR #2174, opened 2026-07-22
chore(templates): remove vestigial wizard icon assets

### LINKED ISSUE #2173: Remove vestigial wizard icon assets from Uno.Templates (.template.config/Assets)
### Summary

The 278 SVG icons under `src/Uno.Templates/content/unoapp/.template.config/Assets/` are **not consumed by any wizard renderer** and should be removed.

### Background

Both wizard renderers — the **VS VSIX wizard** and the **Live Wizard** — are built from `unoplatform/uno.studio`. At build time uno.studio (`ExtractTemplateJson.targets`) downloads the `Uno.Templates` package and extracts **only** the metadata JSON files (`template.json`, `TemplateWizard.json`, `dotnetcli.host.json`). It renders icons from its **own** bundled `Studio.Shared/Assets/`, resolving the `Icon` path in `TemplateWizard.json` against `ms-appx` — never against the template package. `uno.rider` uses text checkboxes (no icons); VS Code has no wizard.

The bundled copies were added in 2024 ("pack needed assets with the templates") but were never wired into a renderer, so they have silently drifted from uno.studio:
- `Features.NativeRendering` / `Features.SkiaRendering` (added to studio ~Apr 2025) were never added here.
- `Platforms.AndroidTV` / `AndroidAuto` / `AndroidWear` were never added here.
- `Platforms.Desktop` and `Template_*` dark variants diverged from uno.studio.

### Proposed change

Remove `.template.config/Assets/` (the SVGs) while keeping the `Icon` path metadata in `TemplateWizard.json` (that is what uno.studio reads to locate each icon in its own asset set). This eliminates the duplication and the whole drift class; new wizard icons then live only in uno.studio.

### Validation

- `dotnet pack` succeeds; packaged nupkg retains the 3 JSONs uno.studio extracts, contains 0 wizard SVGs, and keeps all `Icon` references intact.
- `dotnet new` project generation is unaffected (icons are not used at generation time).
- All wizard `Icon` references resolve to existing assets in uno.studio's `Studio.Shared/Assets/`.

### Type

Refactoring / cleanup (no functional or API change)

agneszitte:
GitHub Issue: closes #2173

## PR Type

- Build or CI related changes

`src/Uno.Templates/content/unoapp/.template.config/Assets/` bundles 278 wizard-icon SVGs (4 theme variants each). These are **not consumed by any wizard renderer**:

- The **VS VSIX wizard** and **Live Wizard** are both built from `unoplatform/uno.studio`. At build time uno.studio (`ExtractTemplateJson.targets`) downloads the `Uno.Templates` package and extracts **only** the metadata JSONs (`template.json`, `TemplateWizard.json`, `dotnetcli.host.json`) — never the `Assets/`. Icons are rendered from uno.studio's own `Studio.Shared/Assets/`, resolving the `Icon` path in `TemplateWizard.json` against `ms-appx`.
- `uno.rider` uses text checkboxes (no icons); VS Code has no wizard.

Because nothing renders them, the bundled copies (added in 2024, "pack needed assets with the templates") drifted from uno.studio: `Features.NativeRendering`/`SkiaRendering` and `Platforms.Android{TV,Auto,Wear}` were never added here, and `Platforms.Desktop` / `Template_*` dark variants diverged.

## What is the new behavior?

The `.template.config/Assets/` SVGs are removed. The `Icon` path metadata in `TemplateWizard.json` is intentionally kept — that is what uno.studio reads to locate each icon in its own asset set. New wizard icons now live only in uno.studio, removing the duplication and the whole drift class.

## PR Checklist

## Other information

Validated locally:
- `dotnet pack` succeeds; the packaged nupkg retains the 3 JSONs uno.studio extracts, contains **0** wizard SVGs, and keeps all 63 `Icon` references intact.
- `dotnet new unoapp/unolib/...` install + generation unaffected (icons are not used at generation).
- All 58 wizard `Icon` references in `TemplateWizard.json` resolve to existing assets in uno.studio's `Studio.Shared/Assets/` (enabled light+dark), so the VSIX wizard and Live Wizard render unchanged.


agneszitte:
## Local validation — VS VSIX wizard & Live Wizard consumption path ✅

Validated the cleaned package end-to-end against **real uno.studio build tooling** (both the VSIX wizard and the Live Wizard share this exact `ExtractTemplateJson` + `Studio.Shared` path):

1. **Package** — `dotnet pack` clean; nupkg keeps the 3 JSONs studio extracts (`template.json`, `TemplateWizard.json`, `dotnetcli.host.json`), contains **0** wizard SVGs, and all **63** `Icon` references intact.
2. **`dotnet new`** — install + `unoapp`/`unolib`/… generation unaffected (icons aren't used at generation).
3. **Studio VSIX packaging** — injected the cleaned package into `VisualStudio.Templates`' download folder and built it: **build succeeded (0/0)**, feed download skipped, produced `Uno.Templates.vsix.nupkg` from the 0-SVG package.
4. **Studio extraction (runtime input)** — ran studio's actual `DownloadTemplateNuGet → BundleTemplates → UnzipTemplate` targets: extracted exactly the 3 `unoapp` metadata JSONs the wizard reads, `TemplateWizard.json` with all 63 `Icon` refs intact, and **0 wizard-icon SVGs** under any `.template.config/Assets/` (the only SVGs present are the generated app's own assets — `dotnet_bot.svg`, splash, app icons).
5. **Icon resolution** — all **58** distinct wizard `Icon` references resolve to existing assets in uno.studio's own `Studio.Shared/Assets/` (enabled light + dark), so the wizard renders unchanged.

Conclusion: removing the bundled templates icons does not affect either wizard — studio extracts only the metadata JSONs from the package and renders icons from its own asset set.

Xiaoy312 (approved):

nickrandolph:
@agneszitte the only issue with this change is that we'd still need to remember to add icons to the vsix. The original intent was to update the vsix to rely on the assets in the template. I'd prefer us to finish the update to the vsix to use the icons in the template. This way as we add/remove features we only need to modify the template, not the vsix

agneszitte:
> @agneszitte the only issue with this change is that we'd still need to remember to add icons to the vsix. The original intent was to update the vsix to rely on the assets in the template. I'd prefer us to finish the update to the vsix to use the icons in the template. This way as we add/remove features we only need to modify the template, not the vsix

@nickrandolph, the Visual Studio VSIX is the only extension + Live Wizard, both in Uno.Studio, currently using these assets. The VS Code extension does not use them because it has no included wizard, and the Rider plugin does not use them either.

In practice, asset-related changes are often duplicated across `uno.templates` and `uno.studio`, sometimes with gaps. The Rider plugin is also frequently forgotten or overlooked and handled as an afterthought when new templates/wizard options are introduced (Example with https://github.com/unoplatform/uno.rider/issues/522 and https://github.com/unoplatform/uno.rider/pull/523 created after I noticed it missing)

In my opinion, these assets should remain in `uno.studio`, where they are actually used. We should then establish a better process, such as instructions, skills, specifications, or CI validation, to detect gaps and verify alignment whenever a new option is added.

Even when assets are added, some can still be missed, as in this example: https://github.com/unoplatform/uno.studio/pull/1059

Overall, I think we need a more reliable way to keep `uno`, `uno.templates`, `uno.studio`, and `uno.rider` aligned when new templates/wizard options are introduced.

nickrandolph:
> > @agneszitte the only issue with this change is that we'd still need to remember to add icons to the vsix. The original intent was to update the vsix to rely on the assets in the template. I'd prefer us to finish the update to the vsix to use the icons in the template. This way as we add/remove features we only need to modify the template, not the vsix
> 
> @nickrandolph, the Visual Studio VSIX is the only extension + Live Wizard, both in Uno.Studio, currently using these assets. The VS Code extension does not use them because it has no included wizard, and the Rider plugin does not use them either.
> 
> In practice, asset-related changes are often duplicated across `uno.templates` and `uno.studio`, sometimes with gaps. The Rider plugin is also frequently forgotten or overlooked and handled as an afterthought when new templates/wizard options are introduced (Example with [unoplatform/uno.rider#522](https://github.com/unoplatform/uno.rider/issues/522) and [unoplatform/uno.rider#523](https://github.com/unoplatform/uno.rider/pull/523) created after I noticed it missing)
> 
> In my opinion, these assets should remain in `uno.studio`, where they are actually used. We should then establish a better process, such as instructions, skills, specifications, or CI validation, to detect gaps and verify alignment whenever a new option is added.
> 
> Even when assets are added, some can still be missed, as in this example: [unoplatform/uno.studio#1059](https://github.com/unoplatform/uno.studio/pull/1059)
> 
> Overall, I think we need a more reliable way to keep `uno`, `uno.templates`, `uno.studio`, and `uno.rider` aligned when new templates/wizard options are introduced.

Right but this is why we moved everything into uno.templates in the first place. Updating the json and assets in uno.template should be all that's required in order to update the vsix an uno.rider (I know it doesn't use the assets but it does use the json file to determine features/options etc). 

The only thing to be said for removing the assets from uno.templates it that it would reduce the size of the package but still leaves the process of adding features/options split across multiple repos, which is what we were trying to avoid. 

At the end of the day, @agneszitte if you'd prefer to remove them from uno.templates and just administer them in uno.studio, it's fine, so long as the process is known.

PR #2215, opened 2026-08-14
fix: Allow MSAL authentication with the desktop platform in the wizard

kazo0:
GitHub Issue (If applicable): none — found while pairing the wizard with unoplatform/uno#24084; happy to file one if preferred.

## PR Type

- Bugfix

The wizard's `authentication.msal` option carries a `platforms` requirement of `!desktop`, so selecting the desktop platform greys MSAL out with *"MSAL only supported on iOS, Android, WebAssembly and Windows App SDK"*.

This restriction is wizard-only and inconsistent with the CLI: `template.json` has no such constraint, and `dotnet new unoapp -auth msal` with desktop selected already generates an app whose desktop head builds and signs in — the loopback flow is pure MSAL.NET, and the Uno.Extensions MSAL provider explicitly supports desktop (its `MsalCacheHelper` token-cache persistence exists *specifically* for desktop; mobile persists natively).

## What is the new behavior?

The `platforms` requirement is removed; `authentication.msal` now only requires Dependency Injection, same as the OIDC and Web options. MSAL can be selected with the desktop platform in the wizard.

## Other information

Pairs with [unoplatform/uno#24084](https://github.com/unoplatform/uno/pull/24084), which fixes the one real desktop gap behind this restriction: the Uno.Sdk not marking desktop as MSAL-supported, so `AuthenticationMsal` apps resolve `Uno.WinUI.MSAL` / `Microsoft.Identity.Client(.Extensions.Msal)` at stale transitive nuspec floors on that head instead of SDK-pinned versions. Best merged once that change ships in an Uno.Sdk referenced here — draft until then.

## PR Checklist

https://claude.ai/code/session_019BJKK7KeJ14DVxxW5UmJqo

### DIFF
--- a/src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
+++ b/src/Uno.Templates/content/unoapp/.template.config/TemplateWizard.json
@@ -420,12 +420,6 @@
 				"dependencyInjection": {
 					"RequiredValues": [ "true" ],
 					"MoreInformation": "Dependency Injection is required"
-				},
-				"platforms": {
-					"RequiredValues": [
-						"!desktop"
-					],
-					"MoreInformation": "MSAL only supported on iOS, Android, WebAssembly and Windows App SDK"
 				}
 			}
 		},
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Package.appxmanifest
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Package.appxmanifest
@@ -22,7 +22,7 @@
       Executable="$targetnametoken$.exe"
       EntryPoint="$targetentrypoint$">
       <uap:VisualElements />
-<!--#if (useProtocolRedirect)-->
+<!--#if (useWebAuthentication || useOidcAuthentication)-->
       <Extensions>
       <uap:Extension Category="windows.protocol">
         <uap:Protocol Name="myprotocol">
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Platforms/Android/MsalActivity.Android.cs
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Platforms/Android/MsalActivity.Android.cs
@@ -8,11 +8,34 @@ using Microsoft.Identity.Client;
 
 namespace MyExtensionsApp._1.Droid;
 
+/// <summary>
+/// Catches the redirect back from the browser after an MSAL sign-in.
+/// </summary>
+/// <remarks>
+/// <para>
+/// MSAL's Android convention is <c>msal{ClientId}://auth</c>, and that is what the MSAL provider
+/// uses unless a <c>RedirectUri</c> is set in the <c>MsalAuthentication</c> configuration section.
+/// Android needs an intent filter matching it, or the browser has nowhere to deliver the code and
+/// the sign-in never completes - it hangs rather than failing.
+/// </para>
+/// <para>
+/// <b><see cref="RedirectScheme"/> must be updated to match your client id.</b> Intent filters are
+/// declared with attributes, which only accept compile-time constants, so this is the one value
+/// that cannot follow appsettings.json automatically.
+/// </para>
+/// </remarks>
 [Activity(NoHistory = true, LaunchMode = LaunchMode.SingleTop, Exported = true)]
 [IntentFilter(
     new[] { Android.Content.Intent.ActionView },
     Categories = new[] { Android.Content.Intent.CategoryDefault, Android.Content.Intent.CategoryBrowsable },
-    DataScheme = "myprotocol")]
+    DataScheme = RedirectScheme,
+    DataHost = "auth")]
 public class MsalActivity : BrowserTabActivity
 {
+    /// <summary>
+    /// "msal" followed by the ClientId from the MsalAuthentication section of appsettings. The
+    /// placeholder below keeps the generated app building; replace it with your own client id, and
+    /// register <c>msal{ClientId}://auth</c> as a redirect URI on the app registration.
+    /// </summary>
+    private const string RedirectScheme = "msal00000000-0000-0000-0000-000000000000";
 }
--- a/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Platforms/iOS/Info.plist
+++ b/src/Uno.Templates/content/unoapp/MyExtensionsApp.1/Platforms/iOS/Info.plist
@@ -40,9 +40,20 @@
 		<false/>
 		-->
 <!--#if (useProtocolRedirect)-->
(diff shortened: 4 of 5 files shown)


kazo0:
Pushed 8f7ebffa here, and surfacing a second finding that bears on this PR's premise.

### Pushed: iOS URL scheme was never registered

`MyExtensionsApp.1/Platforms/iOS/Info.plist` declared `CFBundleURLSchemes` as a **top-level** key. iOS only reads custom URL schemes from [`CFBundleURLTypes`](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleurltypes) — an array of dicts, each carrying its own `CFBundleURLSchemes` — so the key was ignored and `myprotocol` was never registered. The sign-in callback had no way back into the app. Independent of this PR's scope, and it affects every `useProtocolRedirect` auth option: MSAL, OIDC and Web.

### Not pushed: the redirect URI is wrong for desktop and WebAssembly

This one is blocked, so flagging rather than committing.

`appsettings.development.json` emits `"RedirectUri": "myprotocol://auth"` whenever `useMsalAuthentication` is true. The conditional is on the *auth choice*, not on selected platforms, so every head gets that one value — and the four heads have mutually exclusive requirements:

| Head | Requires | Why |
| --- | --- | --- |
| Desktop | `http://localhost` | MSAL runs a loopback listener and rejects anything else. `Microsoft.Identity.Client` carries the error code `loopback_redirect_uri` plus *"Only http uri scheme is supported… Configure http://localhost or http://localhost:port both during app registration and when you create the PublicClientApplication object"* |
| WebAssembly | `{origin}/authentication-callback` | The browser can only land on a real URL on the app's own origin, and Uno's `ICustomWebUi` polls the popup URL for a match — a custom scheme can never appear there. Entra also needs SPA client-type for CORS on the token endpoint |
| Android | `msal{ClientId}://auth` | Custom scheme + matching intent filter |
| iOS | `msauth.{BundleId}://auth` | URL-scheme routing via `CFBundleURLTypes` |

So `myprotocol://auth` works on the two mobile heads (given the manifests and an Entra registration) and cannot work on desktop or WASM. Since this PR makes MSAL **selectable with desktop** in the wizard, that combination would generate an app whose desktop head builds but can't complete sign-in — worth resolving before this leaves draft.

That also softens the claim in the description above that `dotnet new unoapp -auth msal` with desktop "already generates an app whose desktop head builds and signs in". It builds; sign-in should fail on the redirect URI unless `appsettings` was edited by hand. Caveat: my evidence is the MSAL validation path and its error strings, not an observed desktop sign-in from a freshly generated app — worth confirming before the description is updated.

### Why it can't be fixed here yet

There's no per-platform seam in `appsettings` (it varies by *environment*, not target), so no single value can be correct. The fix has to be in the provider, and it exists on `dev/sb/msal-auth-fixes` in `uno.extensions` — `MsalRedirectDefaults` derives the conventional URI per platform, with precedence *platform default < config `RedirectUri` < `Builder(...)` callback*. That file isn't on `uno.extensions` `main` yet, and until it ships, deleting the line would leave MSAL with **no** redirect URI on any head — a regression for the mobile heads that work today.

### Follow-up once that ships

1. Drop `"RedirectUri": "myprotocol://auth"` from `appsettings.development.json` so the platform defaults apply.
2. Retire the three `myprotocol` manifest declarations, which then no longer match anything: the Android `MsalActivity` intent filter (`DataScheme = "myprotocol"` → `msal{ClientId}`, `DataHost = "auth"`), the iOS scheme (→ `msauth.{BundleId}`), and `<uap:Protocol Name="myprotocol">` in `Package.appxmanifest`.
3. Note that the Android scheme can't follow `appsettings` automatically — intent filters are compile-time attributes — so it needs a documented constant or a comment telling the developer to update it.

Context: this surfaced while bringing `Authentication.MsalExtensionsDemo` in `Uno.Samples` up to par with the plain-MSAL sample, where the same stranded `myprotocol` manifests broke Android and iOS after `RedirectUri` was removed from config to exercise the new platform defaults.

https://claude.ai/code/session_01UNhPA6x8uGnNo2YGMNDDVr

kazo0:
Pushed 8b651aeb — the per-platform redirect URI change from the previous comment is now in, so this branch no longer needs that as a follow-up.

**Answering the "is it wired to the bundle id?" question directly: iOS yes, Android no — and Android *can't* be.**

| Platform | Derived URI | Wired to |
| --- | --- | --- |
| iOS | `msauth.{BundleId}://auth` | ✅ the plist carries `msauth.com.companyname.myextensionsapp`, and `appIdEvaluator` substitutes that literal — the same one `<ApplicationId>` in the csproj uses, which is what becomes `CFBundleIdentifier`. Generating as `MsalGen` produced `msauth.com.companyname.msalgen` against `<ApplicationId>com.companyname.msalgen</ApplicationId>` |
| Android | `msal{ClientId}://auth` | ❌ not the application id at all — it's the **Azure client id**, unknown at generation time. Left as a documented placeholder constant, since intent filters are attributes and take compile-time constants only |
| WinAppSDK | broker-managed | n/a — MSAL never redirects through `myprotocol`, so an MSAL app no longer ships that protocol registration |

The Android placeholder is the one manual step left. It generates a valid manifest so the app still builds, and the XML doc on `RedirectScheme` says what to replace it with. Worth deciding whether that's good enough or whether it should be surfaced somewhere more visible than a comment — a wizard field for the client id would let it be substituted like the app id, but that's a bigger change than this PR.

**Web/OIDC are untouched.** `useProtocolRedirect` is shared across all three auth options, so the plist block is now split by auth type: MSAL gets the `msauth.` dict, Web and OIDC keep a `myprotocol` dict. `Package.appxmanifest` is gated on `useWebAuthentication || useOidcAuthentication`.

### Verification

Expanded every conditional branch for all four auth scenarios and validated the resulting manifests (XML well-formedness, expected schemes present/absent, `RedirectUri` gone from the MSAL section, no stray top-level `CFBundleURLSchemes`) — msal / oidc / web / none all pass. Then packed the template locally and generated two apps:

- **MSAL app** — plist has only `msauth.com.companyname.msalgen`; `grep -rn myprotocol` across the whole generated app returns nothing; no `uap:Protocol`; `MsalActivity` has `DataScheme = RedirectScheme` + `DataHost = "auth"`. Android and iOS heads both build clean, and the generated `AndroidManifest.xml` shows `android:scheme="msal00000000-…"` with `android:host="auth"`.
- **OIDC app** — plist keeps `myprotocol`, `uap:Protocol` present, no `MsalActivity` emitted.

One caveat carried over from the earlier comment: I still haven't observed a real end-to-end sign-in on any head from a generated app — the client id placeholder makes that a manual step. What's verified is that the derived URI and the declared manifest scheme now agree, which is what was broken.

https://claude.ai/code/session_01UNhPA6x8uGnNo2YGMNDDVr

PR #2245, opened 2026-09-03
feat!: Move the WinAppSdk group to 2.4.0

MartinZikmund:
GitHub Issue (If applicable): relates to unoplatform/uno-private#2297

## PR Type

- Feature
- Build or CI related changes

The shipped `Uno.Sdk` manifest pins `Microsoft.WindowsAppSDK` at **1.7.250909003**, while `unoplatform/uno` builds and tests against **2.4.0**:

| Project in `unoplatform/uno` (on `master` today) | Pinned |
|---|---|
| `SamplesApp/SamplesApp.csproj:213` | 2.3.1 |
| `Uno.UI.RuntimeTests.Windows.csproj:65` | 2.3.1 |
| `Uno.UI.Extras.Windows.csproj:72` | 2.3.1 |
| `Uno.WinAppSDKSyncGenerator.References.csproj:15` | 2.3.1 |

unoplatform/uno#24378 moves all of those to 2.4.0 as well. So an app built on the Windows head gets a WinAppSDK a **full major behind** the surface Uno is validated against. `WinAppSdk` sits in `.github/sdk-update-exclude.json`, so the updater can never move it — it only changes by hand, which is why the gap has persisted.

## What is the new behavior?

The `WinAppSdk` group moves to **2.4.0**, matching unoplatform/uno#24378 which puts core on the same version uniformly.

**Why 2.4.0.** Core is moving to the same version in unoplatform/uno#24378, which also fixes a `1.0.0` WinAppSDK floor that `Uno.WinUI`'s nuspec has been publishing. Landing both together is what makes the version uniform end to end; landing this one alone would ship customers a version core has not built against, so **these two should merge together**.

`ReadMe.md` is updated in lockstep (the version table row and its embedded copy of the manifest); the embedded manifest was verified byte-identical to `packages.json` after the edit.

## Validation

**Inspection + consistency check only.** No build was run: this is a version-table change whose effect is on generated customer apps, which is what the template test matrix in CI exercises. Treat CI on this PR as the real validation — in particular the Windows test groups.

Worth a reviewer's attention: this is a **major** WinAppSDK move for every consuming app, so the Windows groups going green is the load-bearing signal here.

## PR Checklist

- [ ] Tested code with current supported SDKs
- [ ] Docs have been added/updated
- [ ] Unit Tests and/or UI Tests
- [ ] Wasm UI Tests unaffected
- [x] Associated with an issue (internal)

**Breaking change:** yes, for Windows-head apps — a WinAppSDK major. Intentional for the 7.0 line.

https://claude.ai/code/session_019vRaBaNK2HnyWeg3TBzEu6

### DIFF
--- a/src/Uno.Sdk/packages.json
+++ b/src/Uno.Sdk/packages.json
@@ -121,7 +121,7 @@
   },
   {
     "group": "WinAppSdk",
-    "version": "1.7.250909003",
+    "version": "2.4.0",
     "packages": [
       "Microsoft.WindowsAppSDK"
     ]
--- a/src/Uno.Sdk/ReadMe.md
+++ b/src/Uno.Sdk/ReadMe.md
@@ -17,7 +17,7 @@ The Uno.Sdk powers the Uno Platform Single Project, including the ability to imp
 | UnoResizetizerVersion | 1.13.0-dev.17 |
 | SkiaSharpVersion | 3.119.2 |
 | SvgSkiaVersion | 3.0.6 |
-| WinAppSdkVersion | 1.7.250909003 |
+| WinAppSdkVersion | 2.4.0 |
 | WinAppSdkBuildToolsVersion | 10.0.28000.2705 |
 | WinAppSdkBuildToolsWinAppVersion | 0.6.1 |
 | MicrosoftLoggingVersion** | 9.0.19 |
@@ -168,7 +168,7 @@ The Uno.Sdk powers the Uno Platform Single Project, including the ability to imp
   },
   {
     "group": "WinAppSdk",
-    "version": "1.7.250909003",
+    "version": "2.4.0",
     "packages": [
       "Microsoft.WindowsAppSDK"
     ]


morning4coffe-dev (commented):
The manifest and both README representations agree on 2.4.0, and that matches the current core runtime/test references. This remains a major consumer runtime change. I could not confirm the required generated Windows app matrix from the currently reported checks; those show analysis/conventional/summary results, not the Windows build evidence called load-bearing in the description. Please link or run that validation before final approval. I have not represented the version-table consistency check as an app compatibility test.

morning4coffe-dev:
New CI now provides a concrete native Windows result: [Windows Tests (Group7_net10)](https://github.com/unoplatform/uno.templates/actions/runs/34491404353/job/103072671339) builds generated Issue396 for `net10.0-windows10.0.26100`, but fails with MSB3030 because `HybridWebView.js` is missing from the Microsoft.Maui.Controls.Compatibility, Microsoft.Maui.Controls.Core, and Microsoft.Maui.Core 10.0.0 package paths.

I have not established whether this is introduced by this PR or already fails on its base. The [Group0 successes](https://github.com/unoplatform/uno.templates/actions/runs/34491404353/job/103072671340) I inspected are `net10.0-desktop` builds, so those alone do not validate native WinAppSDK compatibility. Keeping the review comment-only pending passing native Windows validation or a verified base/head comparison.

src/Uno.Sdk/packages.json
```diff
@@ -121,7 +121,7 @@
   },
   {
     "group": "WinAppSdk",
-    "version": "1.7.250909003",
+    "version": "2.4.0",
```
morning4coffe-dev, line 124: [P1] Preserve the existing native Windows MAUI template build

This version change breaks the existing Issue396 scenario (`-maui -markup csharp -platforms windows -platforms android -tfm net10.0`). The [exact-base job at 57cbfe957a5b](https://github.com/unoplatform/uno.templates/actions/runs/33832641333/job/100899986914) builds its `net10.0-windows10.0.26100` head successfully; the [PR job at b5ebe858a2ac](https://github.com/unoplatform/uno.templates/actions/runs/34491404353/job/103072671339) fails with three MSB3030 errors. Both binary logs evaluate .NET SDK 10.0.100, MAUI 10.0.0, and Uno runtime 6.8.0-dev.130.

In the failing binary log, `Microsoft.Windows.SDK.BuildTools.MSIX/1.7.251221100` adds `AddPriPayloadFilesToCopyToOutputDirectoryItems` / `WinAppSdkExpandPriContent`. It emits `Microsoft.Maui/Handlers/HybridWebView/HybridWebView.js` copy inputs under the `lib/net10.0-windows10.0.20348` directories of Microsoft.Maui.Controls.Compatibility, Microsoft.Maui.Controls.Core, and Microsoft.Maui.Core; none of those source files exists. The base build does not execute this expansion target or add those copy inputs.

Please coordinate a compatible SDK/MAUI asset-packaging fix (or a compatible version choice) before enabling the upgrade for generated apps, and keep this scenario enabled in the Windows matrix. Suppressing the copy errors would leave the resource payload missing.

morning4coffe-dev (changes requested):
Updating my earlier comment-only assessment with the now-verified base/head comparison. The exact base's generated native Windows Issue396 build succeeds; this PR's equivalent build fails. The binary logs trace the failure to the new PRI-payload expansion/copy path introduced by the WinAppSDK dependency change, rather than an unclassified CI outage. Details and the two pinned CI jobs are inline. Evidence is from the published CI logs and binary logs, not a new local app run.

PR #2246, opened 2026-09-03
feat!: Sweep the AndroidX and Material bindings to current

MartinZikmund:
GitHub Issue (If applicable): relates to unoplatform/uno-private#2297

## PR Type

- Feature

`UpdateGroup` skips every group whose packages start with `Xamarin`:

```csharp
// tools/Uno.Sdk.Updater/Program.cs:344-350
// Skip AndroidX packages to avoid Java misalignment
else if (group.Packages.Any(x => x.StartsWith("Xamarin")) ...)
    Console.WriteLine($"Skipping '{group.Group}' to avoid Java misalignment.");
```

That skip is correct — these bindings have to move as a coordinated set, not one at a time — but it also means they never move at all unless someone does it deliberately. A major is that moment.

## What is the new behavior?

The `net10.0` override of all 14 AndroidX/Material groups moves to current:

| Group | net10.0 was | now |
|---|---|---|
| AndroidMaterial | 1.12.0.5 | **1.14.0.6** |
| AndroidXAppCompat | 1.7.1.1 | **1.8.0** |
| AndroidXActivity | 1.10.1.3 | **1.13.0.1** |
| AndroidXBrowser | 1.8.0.11 | **1.10.0.1** |
| AndroidXCarApp | 1.4.0.3 | **1.7.0.4** |
| AndroidXCollection | 1.5.0.3 | **1.6.0.1** |
| AndroidXLeanback | 1.0.0.31 | **1.2.0.4** |
| AndroidXLegacySupportV4 | 1.0.0.33 | **1.0.0.37** |
| AndroidXNavigation | 2.9.2.1 | **2.9.8.1** |
| AndroidXRecyclerView | 1.4.0.3 | **1.4.0.6** |
| AndroidXSplashScreen | 1.0.1.14 | **1.2.0.3** |
| AndroidXSwipeRefreshLayout | 1.1.0.29 | **1.2.0.3** |
| AndroidXWear | 1.3.0.15 | **1.4.0.2** |
| AndroidXWearTiles | 1.4.1 | **1.6.2** |

**Only the `net10.0` overrides move.** The base values are the net9 ones and are deliberately untouched.

## Validation

Every package in every group was checked against nuget.org for the target version — not just one probe package per group, since a group is only usable if *all* of its packages publish the same version. All 14 groups have the candidate available for every package they list; zero mismatches.

`ReadMe.md`'s embedded copy of the manifest was regenerated and verified byte-identical to `packages.json`.

**No Android build was run locally.** Java/AndroidX misalignment is exactly the failure this skip list exists to prevent, and it shows up at Android build time, not at restore. **The Android test groups in CI are the real gate for this PR** — if they go red, the right response is to bisect the set rather than to lower individual versions blindly.

## Other information

Related, and *not* fixed here: `PackageManifest.GetGroupVersion` looks the override up by exact TFM key (`VersionOverride.TryGetValue(_targetFrameworkVersion, ...)`), and `TargetFrameworkVersion` is the part before the `-`. Every override in this manifest is keyed `net10.0` only, so a **net11** app silently falls back to the net9-era base version — for these groups and for Maui, WasmBootstrap, MicrosoftLoggingConsole and WindowsCompatibility too. Uno Platform 7.0 ships net11 assemblies, so this becomes real as soon as net11 templates exist. It wants fixing in `unoplatform/uno` (nearest-lower-TFM lookup) rather than by adding a second key to 18 groups here.

## PR Checklist

- [ ] Tested code with current supported SDKs
- [ ] Docs have been added/updated
- [ ] Unit Tests and/or UI Tests
- [ ] Wasm UI Tests unaffected
- [x] Associated with an issue (internal)

**Breaking change:** yes for Android apps — coordinated binding bump. Intentional for the 7.0 line.

https://claude.ai/code/session_019vRaBaNK2HnyWeg3TBzEu6

### DIFF
--- a/src/Uno.Sdk/packages.json
+++ b/src/Uno.Sdk/packages.json
@@ -200,7 +200,7 @@
       "Xamarin.Google.Android.Material"
     ],
     "versionOverride": {
-      "net10.0": "1.12.0.5"
+      "net10.0": "1.14.0.6"
     }
   },
   {
@@ -210,7 +210,7 @@
       "Xamarin.AndroidX.Legacy.Support.V4"
     ],
     "versionOverride": {
-      "net10.0": "1.0.0.33"
+      "net10.0": "1.0.0.37"
     }
   },
   {
@@ -220,7 +220,7 @@
       "Xamarin.AndroidX.Core.SplashScreen"
     ],
     "versionOverride": {
-      "net10.0": "1.0.1.14"
+      "net10.0": "1.2.0.3"
     }
   },
   {
@@ -230,7 +230,7 @@
       "Xamarin.AndroidX.AppCompat"
     ],
     "versionOverride": {
-      "net10.0": "1.7.1.1"
+      "net10.0": "1.8.0"
     }
   },
   {
@@ -240,7 +240,7 @@
       "Xamarin.AndroidX.RecyclerView"
     ],
     "versionOverride": {
-      "net10.0": "1.4.0.3"
+      "net10.0": "1.4.0.6"
     }
   },
   {
@@ -250,7 +250,7 @@
       "Xamarin.AndroidX.Activity"
     ],
     "versionOverride": {
-      "net10.0": "1.10.1.3"
+      "net10.0": "1.13.0.1"
     }
   },
   {
@@ -260,7 +260,7 @@
       "Xamarin.AndroidX.Browser"
     ],
     "versionOverride": {
-      "net10.0": "1.8.0.11"
+      "net10.0": "1.10.0.1"
     }
   },
   {
@@ -270,7 +270,7 @@
       "Xamarin.AndroidX.SwipeRefreshLayout"
     ],
     "versionOverride": {
-      "net10.0": "1.1.0.29"
+      "net10.0": "1.2.0.3"
     }
   },
   {
@@ -280,7 +280,7 @@
       "Xamarin.AndroidX.Leanback"
     ],
     "versionOverride": {
-      "net10.0": "1.0.0.31"
+      "net10.0": "1.2.0.4"
     }
   },
   {
@@ -290,7 +290,7 @@
       "Xamarin.AndroidX.Car.App.App"
     ],
     "versionOverride": {
-      "net10.0": "1.4.0.3"
+      "net10.0": "1.7.0.4"
     }
   },
   {
@@ -300,7 +300,7 @@
       "Xamarin.AndroidX.Wear"
     ],
     "versionOverride": {
-      "net10.0": "1.3.0.15"
+      "net10.0": "1.4.0.2"
     }
   },
   {
@@ -310,7 +310,7 @@
       "Xamarin.AndroidX.Wear.Tiles"
     ],
     "versionOverride": {
-      "net10.0": "1.4.1"
+      "net10.0": "1.6.2"
     }
   },
   {
@@ -323,7 +323,7 @@
       "Xamarin.AndroidX.Navigation.Common"
     ],
     "versionOverride": {
-      "net10.0": "2.9.2.1"
+      "net10.0": "2.9.8.1"
     }
   },
   {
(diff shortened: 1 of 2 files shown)


morning4coffe-dev (commented):
The source consistently updates only the net10.0 overrides, and the README manifest matches the changed JSON. Package availability is not enough to validate the Java/AndroidX compatibility that this coordinated bump depends on. My independent live-feed probe was blocked by local NuGet TLS failures, and the currently reported checks do not show the Android generated-app build/link matrix. Leaving approval pending that platform evidence, rather than treating static checks or the author's availability sweep as proof of Android compatibility. The documented net11 exact-key fallback remains a separate runtime SDK issue.

morning4coffe-dev:
I inspected the new [Build failure](https://github.com/unoplatform/uno.templates/actions/runs/34491398357/job/103115916935) at cce480ef7afd. Uno.Sdk restore stops with NU1301: the SkiaSharp EAP service index returns HTTP 500 with an explicit concurrency/IP-address rate-limit message.

The generated platform matrix was skipped, so the green Tests Validation summary is not Android build/link evidence. This log identifies feed throttling, not a demonstrated AndroidX/Material compatibility regression. Approval remains pending the actual platform validation after the feed failure is cleared.

src/Uno.Sdk/packages.json
```diff
@@ -200,7 +200,7 @@
       "Xamarin.Google.Android.Material"
     ],
     "versionOverride": {
-      "net10.0": "1.12.0.5"
+      "net10.0": "1.14.0.6"
```
morning4coffe-dev, line 203: [P1] Coordinate the Android minimum with the binding sweep

These updated pins resolve AndroidX libraries requiring minSdk 23 (the first manifest-merger failure reports `androidx.tracing`; resolved binding `Xamarin.AndroidX.Tracing.Tracing/2.0.0`). The bundled private SDK is still `6.8.0-dev.130`, whose `Uno.Common.Android.targets:6` defaults `SupportedOSPlatformVersion` to `21.0`.

Fresh generated net10 Android apps therefore fail `AMM0000` in both Skia and native modes, while the exact-base equivalents build successfully. Setting only `-p:SupportedOSPlatformVersion=23.0` produces valid APKs with these same package pins.

Please ship the corresponding net10 Android default/minimum-support change and migration note with this upgrade (or keep an API-21-compatible coordinated set). The API-23 override above is diagnostic, not a fix already in this PR; suppressing the failure with `tools:overrideLibrary` would hide the runtime-compatibility problem.

morning4coffe-dev (changes requested):
The independent Android builds now establish a default-template regression, rather than just an evidence gap.

I packaged Uno.Sdk and Uno.Templates from exact base `57cbfe957a5b` and head `cce480ef7afd`, then generated recommended `net10.0-android` apps for Skia/navigation and native/frame navigation:

| Configuration | Skia | Native |
|---|---|---|
| Exact base, default API 21 minimum | APK build passes | APK build passes |
| PR head, default API 21 minimum | Manifest merge fails: API 21 < 23 | Same failure |
| PR head, explicit `SupportedOSPlatformVersion=23.0` (diagnostic only) | APK build passes | APK build passes |

The SDK still defaults Android to API 21, while the updated AndroidX graph contains AARs requiring API 23. Please coordinate the minimum-version change with this sweep, or retain an API-21-compatible set.

These used .NET SDK 10.0.102 / Android SDK pack 36.1.69, Debug `SignAndroidPackage`, android-arm64, AOT disabled and warnings-as-errors. Generated UI/project sources match base byte-for-byte; Uno.WinUI 6.8.0-dev.130, HotDesign and SkiaSharp pins were unchanged. Local SDK/template version labels were used only to isolate the exact-source packages. The exact binding versions came from Microsoft's dotnet-public mirror after the local NuGet.org TLS failure; none were substituted. APK manifests and DEX entries were checked, but no device execution is claimed.

The earlier CI feed-throttling failure is still an infrastructure issue; this is a separate, reproduced product compatibility failure.

PR #2257, opened 2026-09-15
ci: Reject prerelease pins on the stable release lines

agneszitte:

A stable release branch is cut from the default branch at some point, and the default branch legitimately carries **dev versions**. So a freshly cut `release/stable/*` branch **inherits those dev pins**.

Replacing them is a step on the release checklist: for each group, move to a stable version — the previous stable when nothing changed, or a new / bumped stable when the release needs one. It is a step the member in charge of the release has to remember, on a file most of the diff never touches, and nothing verifies it happened.

It did not happen for 6.7. 
The [initial 6.7 version](https://github.com/unoplatform/private/issues/1105) - `Uno.Sdk` **6.7.18** *and* [6.7 SR1](https://github.com/unoplatform/private/issues/1134) - `Uno.Sdk` **6.7.22** both shipped:

| Group | Pinned | Stable available at the time |
|---|---|---|
| `WasmBootstrap` → `versionOverride[net10.0]` | `10.1.0-dev.214` |  `10.0.96` |
| `Resizetizer` | `1.13.0-dev.17` | `1.12.1` |
| `sdkextras` | `6.4.0-dev.13` | `6.3.1` |

All three from an unbranched `main`. The 6.6 line did perform the step — `a4d00c58`, *"fix: replace dev versions with stable 6.6/6.5 releases"*, which replaced two of these exact literals — so the process works when it is remembered. It just is not enforced, and once missed the pins survived two releases unnoticed.

The updater then compounds it: running unattended on a 6-hourly cron with auto-merge, it keeps the inherited dev groups moving *forward* on the dev line (`10.1.0-dev.214` → `.217`, `1.13.0-dev.17` → `.18`), so the branch drifts further from stable rather than back toward it.

## What is the new behavior?

A required check that fails when a stable release branch carries **any** prerelease pin — turning that checklist item from remembered into enforced.

- Covers **every file the updater writes**, not just the manifest:
  - `packages.json`, including **`versionOverride` entries as well as group versions** — that is where the `WasmBootstrap` pin hid, and a check reading only `version` would have missed the worst of the three.
  - **`Uno.Sdk.Updater.targets` and `Uno.Sdk.Updater.props`**, which record the same pins as MSBuild properties. A dev version ships through those even when the manifest is clean — and `Uno.Sdk.Updater.targets` carries one today (`UnoWasmBootstrapVersionNet10`).
- Also verifies the manifest copy embedded in **`ReadMe.md` still matches `packages.json`**. The updater rewrites both together, so a mismatch means one was hand-edited and the shipped package would disagree with its own documentation.
- **Skips on `main` and `servicing/*`**, where prerelease pins are expected and correct.
- **Gates `build`**, the root of the job graph, so a bad pin short-circuits the build, the sign/publish chain and the entire Linux/Windows/macOS template matrix rather than spending runners on a manifest that must not ship. The check itself costs seconds and is a no-op off the stable release lines.
- Branch resolution **mirrors `env.CheckoutRef`**, so a `workflow_dispatch` against a stable branch is checked rather than skipped.

## Validation

This PR's own CI run against `release/stable/6.7` — the check failed in **6 seconds** and every downstream job was skipped:

```
##[error]src/Uno.Sdk/packages.json pins 3 prerelease version(s) on a stable release branch.
##[error]WasmBootstrap -> versionOverride[net10.0] = 10.1.0-dev.214
##[error]Resizetizer -> version = 1.13.0-dev.17
##[error]sdkextras -> version = 6.4.0-dev.13
##[error]src/Uno.Sdk.Updater.targets pins 1 prerelease version(s) on a stable release branch.
##[error]UnoWasmBootstrapVersionNet10 = 10.1.0-dev.214
OK: all 1 pins in src/Uno.Sdk/Uno.Sdk.Updater.props are stable.
OK: src/Uno.Sdk/ReadMe.md matches src/Uno.Sdk/packages.json.
##[error]Process completed with exit code 1.
```

| Job | Outcome |
|---|---|
| Verify Stable Pins | **failure** (6s) |
| Build, Sign Package, Publish Dev / Uno Production / NuGet.org Production | skipped |
| Generate Test Matrix, Linux / Windows / macOS Tests | skipped |
| Report Missing Dependencies | success (ungated by design) |

It finds precisely the four pins across the two files, and nothing else — the other 58 manifest pins and the props file are correctly left alone.

| Case | Result |
|---|---|
| `release/stable/6.7` as-is | **exit 1**, the 3 pins above |
| same manifest, pins replaced with stables | **exit 0**, "all 61 pins are stable" |
| `main` | **exit 0**, skipped |
| `refs/heads/servicing/6.8` | **exit 0**, skipped |

Pester coverage for the decision logic — **24 tests, all passing** — including the cases that matter for false positives:

- `1.7.250909003` and `10.0.28000.2705` (long stable WinAppSdk build numbers) are **not** prerelease
- build metadata after `+` is ignored, so `6.7.135+1e094106` is stable
- `dev/agzi/release-notes` does **not** match as a release branch
- a dev pin hidden only in a `versionOverride` is found
- a property named `UnoWasmBootstrapVersionNet10` is read (a check for names *ending* in `Version` would miss exactly the pin that started this)
- non-version elements in a `.targets` file are ignored

The gating was verified on the workflow itself: `ci.yml` parses, `build.needs = verify_stable_pins`, `sign.needs = build`, and the template matrix stays rooted behind them — so a failed check skips the whole graph.

Run locally with:

```
pwsh build/Verify-StablePins.ps1 -Branch release/stable/6.7
Invoke-Pester -Path build/Verify-StablePins.Tests.ps1, build/Verify-StablePins.MSBuild.Tests.ps1
```

## Also fixed: Tests Validation no longer passes on a skip

This PR's own run exposed a related hole in the same workflow. `Tests Validation` only failed on `failure`, so a **skipped** matrix counted as a pass — it reported success and printed *"All matrix test jobs passed"* for a run where nothing executed. That is the one assurance a required check must never give, and it is reachable from any upstream gate, not just this one.

The matrix jobs are gated purely on upstream success (`generate-test-matrix` and `build`), so they are never skipped for a legitimate reason such as a doc-only change — a skip always means the tests did not run. `skipped` and `cancelled` are now failures; only `success` passes, and the message names the job and its status.

Verified across the four outcomes: all-success passes; all-skipped, a single failure, and a single cancelled each block.

## Note on merge order

This check **will fail for the backport on `release/stable/6.7` until the three pins are resolved**, which is in progress for [6.7 SR2](https://github.com/unoplatform/private/issues/1137):

- `sdkextras` → `6.3.1` — in #https://github.com/unoplatform/uno.templates/pull/2259
- `Resizetizer` → `1.12.2` — pending unoplatform/uno.resizetizer#383
- `WasmBootstrap` → a stable `10.0.x` — pending unoplatform/Uno.Wasm.Bootstrap#1072

So merge this **after** those land, or expect a red (and correct) check until they do.

## PR Checklist

- [x] Tested — 24 Pester tests, the end-to-end runs above, and this PR's own CI run
- [ ] Docs — N/A
- [x] Contains **NO** breaking changes


agneszitte:
@Mergifyio backport release/stable/6.7

agneszitte:
@Mergifyio backport servicing/6.8

kazo0 (approved):

ajpinedam (approved):

PR #2258, opened 2026-09-15
ci: Stop installing the obsolete Android 'tools' package

agneszitte:

**Every macOS template-test job fails**, on every branch, in the step `Install Android 34`:

```
Warning: Failed to find package 'tools'
Loading package information...
Loading local repository...
##[error]Process completed with exit code 1.
```

That step installs five Android SDK packages, and the first is **`tools`** — the long-retired package Google no longer serves from the SDK repository. `sdkmanager` warns, then exits **1**. The step runs under `bash -e -o pipefail`, so it aborts right there and **none** of the four packages that are actually needed is installed.

The CI log is what pins it down: between the warning and the error there is nothing but that same command's own progress output — no output from `platform-tools`, `platforms;android-34` or either `build-tools`. They never ran.

### Nothing in the repo changed — the runner image did

The nightly on `main` flipped with no commits in between:

| Date | Result |
|---|---|
| 2026-09-15 00:36 | **failure** |
| 2026-09-14 00:39 | success |
| 2026-09-13 00:39 | success |
| 2026-09-12 00:34 | success |

And it hits every branch at once — four unrelated branches within the same hour, all macOS groups, while Linux and Windows stay 24/24 green:

| Branch | macOS failed |
|---|---|
| `sdk/update/dev` | 24/24 |
| `sdk/update/release/6.7` | 24/24 |
| `sdk/update/release/6.6` | 20/20 |
| `dev/agzi/verify-stable-pins` | 24/24 |

The call only kept working while the hosted `macos-15` image happened to carry a cached copy of the obsolete package. The image refreshed and the cache went with it.

## What is the new behavior?

The `tools` install is removed. The four packages that matter are untouched.

## Validation

Reproduced locally against a **clean SDK root**, which gives the same message and exit code as CI:

```
$ sdkmanager --sdk_root=<empty dir> --install 'tools'
Warning: Failed to find package 'tools'
EXIT CODE = 1
```

(A first attempt against the machine's own SDK was inconclusive — a corrupted `package.xml` there made *every* invocation return 1 — hence the clean root.)

`sdkmanager --list` against that same clean root confirms the package is gone while the rest are fine, so removing the line is a complete fix rather than a workaround:

| Package | Availability |
|---|---|
| `tools` | **NOT AVAILABLE** |
| `platform-tools` | available |
| `platforms;android-34` | available |
| `build-tools;35.0.0` | available |
| `build-tools;34.0.0` | available |

Nothing depends on it either: no reference to `$ANDROID_HOME/tools` anywhere in `.github/` or `build/`, and no other Uno repo (`uno`, `uno.resizetizer`, `uno.toolkit.ui`) installs that package at all.

## Needs backporting

`release/stable/6.7` and `release/stable/6.6` carry the identical line in their own copy of this composite action, and `uses: ./.github/actions/ci/...` resolves against the checked-out ref — so fixing `main` does not fix them.

This is currently **blocking 6.7 SR2**: the SR2 manifest PR #2247 lives on `sdk/update/release/6.7`, whose macOS legs are red for exactly this reason.

## PR Checklist

- [x] Tested — reproduced the failure and verified package availability against a clean SDK root
- [ ] Docs — N/A
- [x] Contains **NO** breaking changes


agneszitte:
@Mergifyio backport release/stable/6.7

agneszitte:
@Mergifyio backport servicing/6.8

kazo0 (approved):

